PR #1, opened 2019-09-12
Feature/test client separate

denis-yuen:
* separates out dockstore-client into new repo
* collects tests that mostly focus on running cwl, wdl, nextflow workflows
* can be a place for running (and testing) services on the cli
  * simplified the testing profiles though when services are added, may need to re-complicate things
* also will hold file-plugin-parent and java-wes-client projects for now

## future work

* carried over the same encrypted bundle and much of the same infrastructure, can probably clean-up to the point where the encrypted bundle is not needed https://github.com/dockstore/dockstore/issues/2891
* there are extra tests in GeneralWorkflowIT that intermingle launching with more normal CRUD ops that can remain in the main Dockstore repo
* clean up tests https://github.com/dockstore/dockstore/issues/2890


dockstore-cli-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -149,8 +150,30 @@ private static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebservice
         Application<DockstoreWebserviceConfiguration> application = support.getApplication();
         application.run("db", "drop-all", "--confirm-delete-everything", configPath);
 
-        List<String> migrationList = Arrays.asList("1.3.0.generated", "1.3.1.consistency", "test.confidential1", "1.4.0", "1.5.0", "test.confidential1_1.5.0", "1.6.0", "1.7.0");
+        List<String> migrationList = Arrays.asList("1.3.0.generated", "1.3.1.consistency");
         runMigration(migrationList, application, configPath);
+
+        migrationList = Collections.singletonList("../dockstore-webservice/src/main/resources/migrations.test.confidential1.xml");
```
coverbeck: dockstore-webservice isn't going to be present, is it? This may be covered in your notes already though.

denis-yuen: No notes yet, but in brief:
I left the infrastructure/tooling in place. 
The regular confidential bash tooling will leave the confidential scripts in this this location even though the Maven module will be gone. 

Ideally, the confidential testing bundle won't be needed for dockstore-cli
https://github.com/dockstore/dockstore/issues/2891

PR #4, opened 2019-10-16
Support Cromwell config file in Dockstore config

wshands:
Supports specifying a Cromwell config file in the Cromwell execution command line by using a keyword in the Dockstore config file. I.e. adding this line to the Dockstore config file:

cromwell-config-file: /Users/waltershands/cromwell.conf

will allow Dockstore CLI to generate an execution command line that specifies the Cromwell configuration file:
java -Dconfig.file=/Users/waltershands/cromwell.conf -jar /Users/waltershands/.dockstore/libraries/cromwell-47.jar run --inputs /var/folders/vj/xky2f9pn64q6__y6w2kpjx4m0000gn/T/foo15679799386456966359json /Users/waltershands/Documents/UCSC/dockstore_work/helloworld.wdl

Addresses https://ucsc-cgl.atlassian.net/browse/DOCK-960

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
@@ -7,14 +7,13 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-import com.google.common.collect.Lists;
 import com.google.gson.Gson;
 import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.common.FileProvisioning;
@@ -23,6 +22,7 @@ import io.dockstore.common.WdlBridge;
 import io.github.collaboratory.cwl.CWLClient;
 import org.apache.commons.configuration2.INIConfiguration;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import wdl.draft3.parser.WdlParser;
 
@@ -35,6 +35,7 @@ import static io.dockstore.client.cli.Client.IO_ERROR;
 public class CromwellLauncher extends BaseLauncher {
     protected static final String DEFAULT_CROMWELL_VERSION = "44";
     protected Map<String, List<FileProvisioning.FileInfo>> outputMap;
+    protected String cromwellConfigFile;
 
     public CromwellLauncher(AbstractEntryClient abstractEntryClient, DescriptorLanguage language, boolean script) {
         super(abstractEntryClient, language, script);
@@ -49,6 +50,9 @@ public class CromwellLauncher extends BaseLauncher {
     public void initialize() {
         // initialize cromwell location from ~/.dockstore/config
         INIConfiguration config = Utilities.parseConfig(abstractEntryClient.getConfigFile());
+        cromwellConfigFile = config.getString("cromwell-config-file", "");
+
+
         String cromwellVersion = config.getString("cromwell-version", DEFAULT_CROMWELL_VERSION);
         String cromwellLocation =
                 "https://github.com/broadinstitute/cromwell/releases/download/" + cromwellVersion + "/cromwell-" + cromwellVersion + ".jar";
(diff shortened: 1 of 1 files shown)


agduncan94:
Is there a followup issue for adding this to documentation?

wshands:
Documentation issue https://ucsc-cgl.atlassian.net/browse/SEAB-594

denis-yuen:
> Walt and I discussed this; putting it here for general consideration -- instead of having a specific entry for cromwell-config-file, I think this could be more generic like the cwltool-extra-parameters; that way we don't have to keep adding new entries in the config file to support additional entries.
> 
> The difference though is that the config file is a Java VM setting, whereas other settings are application parameters. I think you could have something like cromwell-vm-options and cromwell-extra-parameters to handle both cases.

I think this is a judgement call. cwltool is loaded with extra parameters for a wide variety of useful options.  Not knowing how many vm options or extra parameters Cromwell has and whether they look useful ...

wshands:
Implemented Charles suggestion for the keywords cromwell-vm-options and cromwell-extra-parameters in the Dockstore config file. There are about 9 extra parameters which could be used. On the one hand they are mostly useful for a specific workflow run so typically a user would not add them to the Dockstore config file, and the --input and --type flags would conflict with the ones Dockstore adds to the command line by default so the user could break the command line, but this does 'future proof' Cromwell command option use...

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
```diff
@@ -81,24 +88,46 @@ public void initialize() {
...
+
+        // Add workflow source file
+        arguments.add(primaryDescriptor.getAbsolutePath());
+
         return arguments;
     }
 
+    private String trimAndPrintInput(String input) {
+        String trimmedInput = input.trim();
+        System.out.println(trimmedInput);
```
coverbeck: 1. This feels like a side effect to me; I would probably not have the println, and then you could just get rid of this method and have `String::trim` in your streams code.
2. Regardless of whether you decide to do that or not, it looks like the output in the console is going to be a list of extra arguments and nothing else, but maybe I'm reading it wrong. Could you attach a screenshot of what the output looks like?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
```diff
@@ -49,6 +52,10 @@ public CromwellLauncher(AbstractEntryClient abstractEntryClient, DescriptorLangu
     public void initialize() {
         // initialize cromwell location from ~/.dockstore/config
         INIConfiguration config = Utilities.parseConfig(abstractEntryClient.getConfigFile());
+        cromwellVmOptions = (List)(config.getList("cromwell-vm-options"));
```
coverbeck: If cromwell-vm-options does not exist in the config file, won't this return null? If so, then your code on line 94 will fail with an NPE.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
```diff
@@ -49,6 +52,10 @@ public CromwellLauncher(AbstractEntryClient abstractEntryClient, DescriptorLangu
...
+        cromwellExtraParameters = (List)(config.getList("cromwell-extra-parameters"));
```
coverbeck: Same as previous comment

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
wshands: Removed the method and replaced it with String::trim

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
wshands: Does not return null if no cromwell-vm-options in config; but returns an ArrayList of size= 0.
So code is OK

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
wshands: Returns an ArrayList of size = 0.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
denis-yuen: Is this cast necessary? 
https://commons.apache.org/proper/commons-configuration/apidocs/org/apache/commons/configuration2/AbstractConfiguration.html#getList-java.lang.Class-java.lang.String-

wshands: Yes otherwise there is an incompatible types error: List<java.lang.string> vs List<java.lang.object>

garyluu:
> I still feel there should be a unit test as I suggested in an earlier comment, but all others have approved and it is a fairly small chunk of code, so approving.

Peer pressure!

PR #15, opened 2020-02-05
Feature/2906/singularity test

emlys:
Added integration tests using Singularity to run the md5sum-checker wdl and cwl workflows


dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
```diff
@@ -0,0 +1,140 @@
...
+        // manually register the CWL version of the md5sum-checker workflow locally
+        testingPostgres.runUpdateStatement("update enduser set isadmin = 't' where username = 'DockstoreTestUser2';");
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        Workflow workflow = workflowApi.manualRegister(SourceControl.GITHUB.getFriendlyName(), "DockstoreTestUser2/md5sum-checker",
+                "/checker-workflow-wrapping-workflow.cwl", "test", "cwl", null);
+        workflowApi.refresh(workflow.getId());
+
+        // start saving the output so as to make sure that Singularity actually runs instead of Docker
+        PrintStream old = System.out;  // save the regular output stream so it can be reset later
```
denis-yuen: We use a System Rule for this kind of thing, there should be examples that look somewhat like this https://stefanbirkner.github.io/system-rules/#SystemErrAndOutRule

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
```diff
@@ -0,0 +1,140 @@
...
+            File configTemplate = new File(ResourceHelpers.resourceFilePath("config_for_singularity"));
+            FileUtils.copyFile(configTemplate, tmpFile);
+
+            // append the new cromwell option to the tmp dockstore config
+            FileUtils.writeStringToFile(tmpFile, "\ncromwell-vm-options: -Dconfig.file=" + cromwellConfPath,
+                    StandardCharsets.UTF_8, true);
+            return tmpFile;
+
+        } catch (IOException e) {
+            e.printStackTrace();
```
denis-yuen: In tests, you usually don't catch exceptions 
(i.e. we want the test to fail if there are problems)

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
```diff
@@ -0,0 +1,140 @@
...
+    public void runWDLWorkflow() {
+        // manually register the WDL version of the md5sum-checker workflow locally
+        testingPostgres.runUpdateStatement("update enduser set isadmin = 't' where username = 'DockstoreTestUser2';");
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        Workflow workflow = workflowApi.manualRegister(SourceControl.GITHUB.getFriendlyName(), "DockstoreTestUser2/md5sum-checker",
+                "/checker-workflow-wrapping-workflow.wdl", "test", "wdl", null);
+        workflowApi.refresh(workflow.getId());
+
+        // make a tmp copy of the dockstore config and add to it the cromwell conf file option
```
denis-yuen: It's not clear to me, what location varies?

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowRegressionIT.java
```diff
@@ -83,7 +83,7 @@ public static void getOldDockstoreClient() throws IOException {
         FileUtils.copyURLToFile(url, dockstore);
         assertTrue(dockstore.setExecutable(true));
         url = new URL("https://raw.githubusercontent.com/DockstoreTestUser2/md5sum-checker/1.6.0/checker-input-cwl.json");
-        md5sumJson = temporaryFolder.newFile("md5sum-wrapper-tool.json");
+        md5sumJson = temporaryFolder.newFile("md5sum_cwl.json");
```
denis-yuen: Is this change intentional? 
It doesn't seem obviously related to singularity testing, but I could be wrong.

dockstore-cli-integration-testing/src/test/resources/config_for_singularity
```diff
@@ -0,0 +1,6 @@
+token: 238e064a9e43328b75023263d9439d880fac5829223fe7a918be69b48484f34d
```
denis-yuen: where is this token from? 
It looks like a dockstore token, so probably not concerning but let's double-check and/or comment

scripts/install-tests.sh
```diff
@@ -19,6 +19,41 @@ else
...
+    tar -xzf singularity-${VERSION}.tar.gz && \
+    cd ./singularity && \
+    ./mconfig
+
+    # Compile singularity
+    ./mconfig && \
+    make -C ./builddir && \
+    sudo make -C ./builddir install
+
+    singularity --version
```
denis-yuen: How long does this compilation take? 
It might be worth considering caching this https://docs.travis-ci.com/user/caching/

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowRegressionIT.java
emlys: Yes, I changed it because that json is input for the CWL version of the md5sum checker. I needed to make a second json file to use with the WDL version, so it seemed to make more sense to call them md5sum_cwl.json and md5sum_wdl.json, to show that they are variations of the same thing

dockstore-cli-integration-testing/src/test/resources/config_for_singularity
emlys: copied from the already existing config_file2.txt

scripts/install-tests.sh
emlys: Around 5 minutes for the whole install. will look into that

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
emlys: in the dockstore config file for this test, there needs to be a line:
`cromwell-vm-options: -Dconfig.file=<absolute path to cromwell conf file, which tells cromwell to run with singularity instead of docker>`. That absolute path is different if it's running on travis or locally, so in other places, `ResourceHelpers.resourceFilePath()` is used to find it. But since the path needs to be in the config I had to write it to a tmp copy of the config.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
emlys: ah ok. I guess I was thinking that an IOException from it failing to create or write to the file isn't the kind of error that's relevant to the test. But it would just fail later if the file isn't created, so I'll remove the catch block

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
```diff
@@ -0,0 +1,140 @@
...
+        // start saving the output so as to make sure that Singularity actually runs instead of Docker
+        PrintStream old = System.out;  // save the regular output stream so it can be reset later
+        TeeOutputStream teeOut = new TeeOutputStream(System.out, outContent);
+        PrintStream out = new PrintStream(teeOut, true);  // also print it to the screen
+        System.setOut(out);
+
+        // run the md5sum-checker workflow
+        Client.main(new String[] {
+            "--config",  // this config file passes the --singularity option to cwltool
+            ResourceHelpers.resourceFilePath("config_for_singularity"),
```
coverbeck: Suggest static final variable for config_for_singularity

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
denis-yuen: Ok, thanks. I think this comment can be added to the Javadoc/comment for `generateCromwellConfig`, particularly the fact that the absolute path is needed and changing between travis or locally. Thanks!

dockstore-cli-integration-testing/src/test/resources/config_for_singularity
denis-yuen: Ok, I don't think `config_file2.txt` is normally checked-in i.e. it is part of the confidential bundle. 
That said, this token isn't live in any way (i.e. it doesn't seem to exist in the production DB) so I think this is fine.

scripts/install-tests.sh
emlys: I tried caching the singularity executable, but it has a bunch of dependencies which also had to be cached, and even after that I'm getting an error that I don't see any way to fix. Not sure this is possible or worth the effort.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/SingularityIT.java
```diff
@@ -28,7 +27,10 @@
     @ClassRule
     public static TemporaryFolder temporaryFolder = new TemporaryFolder();
 
-    private final ByteArrayOutputStream outContent = new ByteArrayOutputStream();
+    private static final String CONFIG = ResourceHelpers.resourceFilePath("config_for_singularity");
```
coverbeck: I would name the variable SINGULARITY_CONFIG or SINGULARITY_CONFIG_TEMPLATE or something like that; avoids any confusion further down when it is used.

.travis.yml
```diff
@@ -50,7 +50,12 @@ jobs:
       env:
         - TESTING_PROFILE=regression-integration-tests
       if: branch = master OR branch =~ /^release.*$/
-        
+
+    - stage: singularity-tests
```
denis-yuen: This is getting is own stage which means that Travis-CI will wait for all four coverage test jobs to complete before starting this job. You probably want to run this in parallel with the other coverage jobs since this is invoking the coverage profile anyway. 
Screenshot from 2020-02-12 17-33-11

dockstore-cli-integration-testing/src/test/resources/config_for_singularity
agduncan94: Should the token be regenerated and this change to a conf bundle test?

PR #18, opened 2020-02-28
Feature/star fix

garyluu:
For dockstore/dockstore#3286


dockstore-client/src/main/java/io/dockstore/client/cli/SwaggerUtility.java
```diff
@@ -71,6 +70,7 @@ public static void unzipFile(File zipFile, File unzipDirectory, boolean deleteZi
     /**
      * These serialization/deserialization hacks should not be necessary.
      * Why does this version of codegen remove the setters?
+     * Probably because someone dun goof'd the restful implementation of publish
```
denis-yuen: I don't get this reference

PR #21, opened 2020-04-22
Changes to support Illumina WES endpoints

wshands:
Addresses:
Evaluate Illumina WES endpoint with Dockstore CLI
https://ucsc-cgl.atlassian.net/browse/SEAB-862
Disable WES workflow launch for local entry
https://ucsc-cgl.atlassian.net/browse/SEAB-1082

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
@@ -76,13 +76,14 @@ public class WESLauncher extends BaseLauncher {
     }
 
     protected void addFilesToWorkflowAttachment(List<File> workflowAttachment, File zippedEntry, File tempDir) {
-        try {
-            SwaggerUtility.unzipFile(zippedEntry, tempDir);
-        } catch (IOException e) {
-            exceptionMessage(e, "Unable to get workflow attachment files from zip file " + zippedEntry.getName()
-                    + " Request not sent.", IO_ERROR);
+        if (zippedEntry != null) {
+            try {
+                SwaggerUtility.unzipFile(zippedEntry, tempDir);
+            } catch (IOException e) {
+                exceptionMessage(e, "Unable to get workflow attachment files from zip file " + zippedEntry.getName() + " Request not sent.",
+                        IO_ERROR);
+            }
         }
-
         try {
 
             // A null fileFilter causes all files to be returned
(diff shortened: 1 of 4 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -76,24 +72,18 @@ public void provisionOutputFiles(String stdout, String stderr, String wdlOutputT
     }
 
     protected void addFilesToWorkflowAttachment(List<File> workflowAttachment, File zippedEntry, File tempDir) {
-        try {
-            SwaggerUtility.unzipFile(zippedEntry, tempDir);
-        } catch (IOException e) {
-            exceptionMessage(e, "Unable to get workflow attachment files from zip file " + zippedEntry.getName()
-                    + " Request not sent.", IO_ERROR);
+        if (zippedEntry != null) {
```
coverbeck, line 75: What if zippedEntry is null? You just continue on; is that the intent?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -106,45 +96,24 @@ protected void addFilesToWorkflowAttachment(List<File> workflowAttachment, File
...
             workflowTypeVersion = "v" + WORKFLOW_TYPE_VERSION;
-        } else if ("WDL".equalsIgnoreCase(languageType)) {
-            workflowTypeVersion = WDL_WORKFLOW_TYPE_VERSION;
         }
 
+        String workflowURL = localPrimaryDescriptorFile.getName();
+
+        final File tempDir = Files.createTempDir();
+        List<File> workflowAttachment = new ArrayList<>();
+        addFilesToWorkflowAttachment(workflowAttachment, zippedEntry, tempDir);
```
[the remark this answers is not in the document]
wshands: I fixed this as you suggested

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: Yes, if by some chance no descriptor files or input json were obtained, then the workflow attachment will be empty, and the wes command will be send with an empty workflow attachment. However I suppose there should always be at least the primary descriptor and input json.

PR #22, opened 2020-04-24
Improve help for WES launch command

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/DOCK-1330

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -1307,7 +1307,7 @@ public abstract class AbstractEntryClient<T> {
     private void printWesLaunchHelpBody() {
         out("");
         out("Description:");
-        out("  Launch an entry locally.");
+        out("  Launch an entry on a WES endpoint.");
         out("");
         out("Required parameters:");
         out("  --entry <entry>                     Complete entry path in Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow:develop)");
@@ -1315,7 +1315,7 @@ public abstract class AbstractEntryClient<T> {
         out("Optional parameters:");
         out("  --json <json file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs");
         out("  --yaml <yaml file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs");
-        out("  --descriptor <descriptor type>      Descriptor type used to launch workflow. Defaults to " + CWL.getLowerShortName());
+        out("  --descriptor <descriptor type>      Descriptor type used to launch workflow. Defaults to " + CWL.getShortName());
         out("  --uuid                              Allows you to specify a uuid for 3rd party notifications");
         out("");
     }
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -1304,11 +1304,27 @@ public abstract class AbstractEntryClient<T> {
         printHelpFooter();
     }
 
+    private void printWesLaunchHelpBody() {
+        out("");
+        out("Description:");
+        out("  Launch an entry locally.");
+        out("");
+        out("Required parameters:");
+        out("  --entry <entry>                     Complete entry path in Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow:develop)");
+        out("");
+        out("Optional parameters:");
+        out("  --json <json file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs");
+        out("  --yaml <yaml file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs");
+        out("  --descriptor <descriptor type>      Descriptor type used to launch workflow. Defaults to " + CWL.getLowerShortName());
+        out("  --uuid                              Allows you to specify a uuid for 3rd party notifications");
+        out("");
+    }
+
     private void wesLaunchHelp() {
         printHelpHeader();
         out("Usage: dockstore " + getEntryType().toLowerCase() + " wes launch --help");
         out("       dockstore " + getEntryType().toLowerCase() + " wes launch [parameters]");
-        printLaunchHelpBody();
+        printWesLaunchHelpBody();
         printWesHelpFooter();
         printHelpFooter();
     }
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
@@ -30,7 +30,7 @@ public class WESLauncher extends BaseLauncher {
 
     public WESLauncher(AbstractEntryClient abstractEntryClient, DescriptorLanguage language, boolean script) {
         super(abstractEntryClient, language, script);
-        setLauncherName("wes");
+        setLauncherName("WES");
     }
 
     /**
@@ -91,7 +91,7 @@ public class WESLauncher extends BaseLauncher {
             }
         } catch (Exception e) {
             exceptionMessage(e, "Unable to traverse directory " + tempDir.getName() + " to get workflow "
-                    + "attachment files", GENERIC_ERROR);
+                    + "attachment files.", GENERIC_ERROR);
         }
         return workflowAttachment;
     }
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
@@ -134,7 +134,6 @@ public class WESLauncher extends BaseLauncher {
      */
     private void wesCommandSuggestions(String runId) {
         String wesEntryType = abstractEntryClient.getEntryType().toString().toLowerCase();
-        out("Useful commands for working with WES runs:");
         out("");
         out("To get status for this run: dockstore " + wesEntryType + " wes status --id " + runId
                 + " [--verbose][--wes-url <WES URL>][--wes-auth <auth>]");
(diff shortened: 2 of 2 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -125,4 +127,16 @@ public void runWESCommand(String jsonInputFilePath, File localPrimaryDescriptorF
...
     }
+
+    /**
+     * help text output
+     */
+    private void wesCommandSuggestions() {
+        String wesEntryType = abstractEntryClient.getEntryType().toString().toLowerCase();
+        out("Useful commands for working with WES runs:");
+        out("");
+        out("To get status for a run: dockstore " + wesEntryType + " wes status --id <run id> [--verbose][--wes-url <WES URL>][--wes-auth <auth>]");
```
coverbeck, line 138: My idea was that you put the actual id in there so you can just copy and paste it directly, rather than copying that, then copying the id, then combine them together by editing a line... Something like:
```
To get the status of this run: dockstore wes status --id <the id from line  117/118>. 
```

You may have to word it more carefully to handle extra parameters, but presumably a user is going to want to check on the status, so just give them the exact command to run, if possible.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -125,4 +127,16 @@ public void runWESCommand(String jsonInputFilePath, File localPrimaryDescriptorF
...
+        out("To cancel a run: dockstore " + wesEntryType + " wes cancel --id <run id> [--wes-url <WES URL>][--wes-auth <auth>]");
```
coverbeck, line 139: Same as previous comment -- put the actual id in there.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: Added the run id to the message

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: Added the run id to the message

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -125,4 +128,18 @@ public void runWESCommand(String jsonInputFilePath, File localPrimaryDescriptorF
...
     }
+
+    /**
+     * help text output
+     */
+    private void wesCommandSuggestions(String runId) {
+        String wesEntryType = abstractEntryClient.getEntryType().toString().toLowerCase();
+        out("Useful commands for working with WES runs:");
+        out("");
+        out("To get status for this run: dockstore " + wesEntryType + " wes status --id " + runId
```
coverbeck, line 139: Is the value of `wesEntryType` something like CWL or WDL? Peeked briefly at swagger, and I don't see such a parameter for this endpoint.

If I'm wrong, ignore. If I'm right:

1. Remove it from message
2. Remove from command/help

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -114,7 +115,9 @@ public void runWESCommand(String jsonInputFilePath, File localPrimaryDescriptorF
         try {
             RunId response = clientWorkflowExecutionServiceApi.runWorkflow(workflowParams, languageType, workflowTypeVersion, TAGS,
                     "{}", workflowURL, workflowAttachment);
-            System.out.println("Launched WES run with id: " + response.toString());
+            String runID = response.getRunId();
+            System.out.println("Launched WES run with id: " + runID);
```
coverbeck, line 119: Here you use `System.out.println`, starting on line 137 you use `out` -- should be consistent.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: wesEntryType is either 'workflow' or 'tool'. We don't support tool, and the option to use 'tool' is short circuited earlier in the control flow, so it will never be 'tool' unless someday we do support 'tool'

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1304,11 +1304,27 @@ private void wesHelp() {
         printHelpFooter();
     }
 
+    private void printWesLaunchHelpBody() {
+        out("");
+        out("Description:");
+        out("  Launch an entry locally.");
```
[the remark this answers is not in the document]
wshands: The original text says this; it confused me at first; I think it means launch an entry that is on GitHub but do it locally from command line on your machine? What would be more clear?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1304,11 +1304,27 @@ private void wesHelp() {
...
+        out("");
+        out("Required parameters:");
+        out("  --entry <entry>                     Complete entry path in Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow:develop)");
+        out("");
+        out("Optional parameters:");
+        out("  --json <json file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs");
+        out("  --yaml <yaml file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs");
+        out("  --descriptor <descriptor type>      Descriptor type used to launch workflow. Defaults to " + CWL.getLowerShortName());
```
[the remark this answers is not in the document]
wshands: What should we use instead?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
coverbeck: Ah, my bad... That said, it still seems sort of unnecessary, unless there will ever be a day where:

1. We support tools, AND
2. Specifying tool vs workflow as a command-line argument will return a different result for the same run id.

On the other hand, if this breaks the symmetry of how commands work, then maybe we need it, but it feels odd to force the user to supply an argument that that will never have any impact (unless I'm wrong in my second point).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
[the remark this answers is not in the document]
garyluu: getShortName()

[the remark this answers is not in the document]
wshands: This is what is used in other parts of the file

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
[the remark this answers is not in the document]
garyluu: To me, I would've thought WES implies that it's launched remotely on a WES endpoint

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: Yes it fits will how our command line is structured to launch tools or workflows. I suppose  we could just hardcode 'workflow', but I will just leave it for now.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: now using out only

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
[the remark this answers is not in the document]
wshands: Yes that's what I meant; I fixed the description.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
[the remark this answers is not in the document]
wshands: I changed it to getShortName

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
coverbeck: To be clear, I was not arguing to drop `workflow` everywhere. It makes sense to have it for launch, even if we don't yet support tools. But it adds no value and just gets in the way on the status or cancel commands, which only require the id.

I guess we can circle back to if we take this feature out of beta.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -114,7 +115,9 @@ public void runWESCommand(String jsonInputFilePath, File localPrimaryDescriptorF
         try {
             RunId response = clientWorkflowExecutionServiceApi.runWorkflow(workflowParams, languageType, workflowTypeVersion, TAGS,
                     "{}", workflowURL, workflowAttachment);
-            System.out.println("Launched WES run with id: " + response.toString());
+            String runID = response.getRunId();
+            out("Launched WES run with id: " + runID);
```
coverbeck, line 119: I feel this phrasing could be better, but will defer to others. Maybe something like, which I'm not that wild about:

```
Your WES run has been submitted. Its id is <id>.
```

I don't like `run` and would rather call it a job, but looks like WES calls them runs too, so never mind that part.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -125,4 +128,18 @@ public void runWESCommand(String jsonInputFilePath, File localPrimaryDescriptorF
             }
         }
     }
+
+    /**
+     * help text output
+     */
+    private void wesCommandSuggestions(String runId) {
+        String wesEntryType = abstractEntryClient.getEntryType().toString().toLowerCase();
+        out("Useful commands for working with WES runs:");
```
coverbeck, line 137: Since you have the specific commands below, I would just delete this.

wshands: done

PR #23, opened 2020-04-28
Make test parameter flags required

garyluu:
For dockstore/dockstore#1971

Clarify that one of the flags are required.  Also mention that only CWL supports YAML

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/JCommanderUtility.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/JCommanderUtility.java
@@ -51,10 +51,14 @@ public final class JCommanderUtility {
                 + "  --entry <entry>                     Complete workflow path in Dockstore (ex. NCI-GDC/gdc-dnaseq-cwl/GDC_DNASeq:master)\n"
                 + "   OR\n"
                 + "  --local-entry <local-entry>         Allows you to specify a full path to a local descriptor instead of an entry path\n");
+        out("");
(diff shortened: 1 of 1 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/JCommanderUtility.java
```diff
@@ -51,10 +51,14 @@ public static void printJCommanderHelpLaunch(JCommander jc, String programName,
                 + "  --entry <entry>                     Complete workflow path in Dockstore (ex. NCI-GDC/gdc-dnaseq-cwl/GDC_DNASeq:master)\n"
                 + "   OR\n"
                 + "  --local-entry <local-entry>         Allows you to specify a full path to a local descriptor instead of an entry path\n");
+        out("");
+        out("  --json <json file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs");
+        out("   OR");
+        out("  --yaml <yaml file>                  Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs (Only for CWL)");
```
coverbeck, line 57: It looks WDLHandler [supports yaml](https://github.com/dockstore/dockstore-cli/blob/9a1a51573395b127d8edf566660f386c22ee8e34/dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java#L83). Or do you mean just the array of maps portion is only CWL?

Also, I think `Only` should be `only` here and on the next line.

garyluu: i tested a simple workflow with cromwell yaml, didn't work out so well.

PR #24, opened 2020-04-30
Support DNAStack WES endpoints

wshands:
Changes to support DNAStack WES endpoints
I.e. set correct content types

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
@@ -22,25 +22,24 @@ public class ApiClientExtended extends ApiClient {
      * @param key
      * @return
      */
-    private MediaType getMediaType(String key) {
-        Optional<String> optionalKey = Optional.ofNullable(key);
+    private Optional<MediaType> getMediaType(Optional<String> key) {
         // if key happens to be null or we don't recognize the key then return null
-        MediaType mediaType = null;
-        if (optionalKey.isPresent()) {
-            String keyWESParam = key.toLowerCase();
+        Optional<MediaType> mediaType = Optional.empty();
+        if (key.isPresent()) {
+            String keyWESParam = key.get().toLowerCase();
             switch (keyWESParam) {
             case "workflow_params":
             case "tags":
             case "workflow_engine_parameters":
-                mediaType =  MediaType.APPLICATION_JSON_TYPE;
+                mediaType =  Optional.of(MediaType.APPLICATION_JSON_TYPE);
                 break;
             case "workflow_attachment":
-                mediaType =  MediaType.APPLICATION_OCTET_STREAM_TYPE;
+                mediaType =  Optional.of(MediaType.APPLICATION_OCTET_STREAM_TYPE);
                 break;
             case "workflow_url":
             case "workflow_type":
             case "workflow_type_version":
-                mediaType =  MediaType.TEXT_PLAIN_TYPE;
+                mediaType =  Optional.of(MediaType.TEXT_PLAIN_TYPE);
                 break;
             default:
             }
@@ -50,7 +49,7 @@ public class ApiClientExtended extends ApiClient {
 
 
     public void createBodyPart(MultiPart multiPart, String key, Object formObject) {
-        Optional<MediaType> mediaType = Optional.ofNullable(getMediaType(key));
+        Optional<MediaType> mediaType = getMediaType(Optional.ofNullable(key));
         if (formObject instanceof File) {
             File file = (File)formObject;
             FormDataContentDisposition contentDisp = FormDataContentDisposition.name(key)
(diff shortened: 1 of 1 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -16,16 +17,56 @@
 
 public class ApiClientExtended extends ApiClient {
 
+    /**
+     *
+     * @param key
+     * @return
+     */
+    private MediaType getMediaType(String key) {
```
coverbeck, line 25: 1. I would return an `Optional<MediaType>`.  If a method can return null, then it should return an Optional instead. You end up putting the return value into an Optional anyway, on line 53, so you might as well make it return an Optional here.
2. The parameter `key` should probably  be an `Optional<String>`, which I think is preferred. If you'd rather not do that, I feel you should just do a null check, instead of creating an `Optional` on the next line for the only purpose of calling `isPresent` 3 lines later.

wshands: implemented the suggestions

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -16,16 +17,55 @@
 
 public class ApiClientExtended extends ApiClient {
 
+    /**
+     *
+     * @param key
+     * @return
+     */
+    private Optional<MediaType> getMediaType(Optional<String> key) {
+        // if key happens to be null or we don't recognize the key then return null
```
coverbeck, line 26: This comment is no longer correct now that you have Optional

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -16,16 +17,55 @@
...
+        Optional<MediaType> mediaType = Optional.empty();
+        if (key.isPresent()) {
+            String keyWESParam = key.get().toLowerCase();
+            switch (keyWESParam) {
+            case "workflow_params":
```
coverbeck, line 31: These strings seem fragile, but it doesn't look like there is an easy way to deal with this? What if a new parameter gets added - we'll need to remember to update this.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -16,16 +17,55 @@
...
+        Optional<MediaType> mediaType = getMediaType(Optional.ofNullable(key));
         if (formObject instanceof File) {
             File file = (File)formObject;
             FormDataContentDisposition contentDisp = FormDataContentDisposition.name(key)
                     .fileName(file.getName()).size(file.length()).build();
-            multiPart.bodyPart(new FormDataBodyPart(contentDisp, file, MediaType.APPLICATION_OCTET_STREAM_TYPE));
+            if (mediaType.isPresent()) {
+                multiPart.bodyPart(new FormDataBodyPart(contentDisp, file, mediaType.get()));
+            } else {
+                multiPart.bodyPart(new FormDataBodyPart(contentDisp, file, MediaType.APPLICATION_OCTET_STREAM_TYPE));
```
garyluu, line 60: mediaType.orElseGet(MediaType.APPLICATION_OCTET_STREAM_TYPE);

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -16,16 +17,55 @@
...
     public void createBodyPart(MultiPart multiPart, String key, Object formObject) {
+        Optional<MediaType> mediaType = getMediaType(Optional.ofNullable(key));
         if (formObject instanceof File) {
             File file = (File)formObject;
             FormDataContentDisposition contentDisp = FormDataContentDisposition.name(key)
                     .fileName(file.getName()).size(file.length()).build();
-            multiPart.bodyPart(new FormDataBodyPart(contentDisp, file, MediaType.APPLICATION_OCTET_STREAM_TYPE));
+            if (mediaType.isPresent()) {
+                multiPart.bodyPart(new FormDataBodyPart(contentDisp, file, mediaType.get()));
+            } else {
```
denis-yuen, line 59: Could the `else` here and on line 66 be redundant if 57 and 64 do `return`

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
denis-yuen: This is even better

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
wshands: removed comment

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
wshands: Used the suggestion and fixed code

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
wshands: Gary's suggestion resulted in removing if else

PR #25, opened 2020-05-12
Obtain Dockstore command line script and jar from dockstore-cli repo

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/DOCK-1334

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -104,6 +104,7 @@ public class Client {
     public static final AtomicBoolean INFO = new AtomicBoolean(false);
     public static final AtomicBoolean SCRIPT = new AtomicBoolean(false);
     public static final String DEPRECATED_PORT_MESSAGE = "Dockstore webservice has deprecated port 8443 and may remove it without warning. Please use 'https://dockstore.org/api' via the Dockstore config file (\"~/.dockstore/config\" by default) instead.";
+    public static final String DOCKSTORE_CLI_REPO_URL = "https://api.github.com/repos/dockstore/dockstore-cli/releases";
 
     private static final Logger LOG = LoggerFactory.getLogger(Client.class);
     private static ObjectMapper objectMapper;
@@ -193,7 +194,8 @@ public class Client {
         URL url;
         try {
             ObjectMapper mapper = getObjectMapper();
-            url = new URL("https://api.github.com/repos/dockstore/dockstore/releases");
+            //url = new URL("https://api.github.com/repos/dockstore/dockstore/releases");
+            url = new URL(DOCKSTORE_CLI_REPO_URL);
             List<Map<String, Object>> mapRel;
             try {
                 TypeFactory typeFactory = mapper.getTypeFactory();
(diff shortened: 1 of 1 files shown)


wshands:
Currently the CLI jar looks on the Dockstore webservice GitHub repo at the releases to find which is the latest or a prerelease; but should actually be looking at the dockstore-cli repo to find out what the dockstore-cli releases are. So now it thinks the latest CLI release is 1.8.5 when the dockstore-cli release is really 1.8.0.

When the CLI jar locates the desired release or prerelease it download the 'dockstore' script from that release. The 'dockstore' script is then responsible for downloading the correct CLI JAR.

So I change the paths in CLI jar to point to dockstore-cli but then it wants to download the 'dockstore' script from the dockstore-cli repo but it isn't there. So we put the 'dockstore' script on the Dockstore CLI release page.

There is a problem for a user that has an old CLI JAR that is still looking on the webservice release page to find latest or prerelease versions.

Here is a potential solution:

- Keep the 'dockstore' script on dockstore/dockstore and also on dockstore/dockstore-cli. 

- Make a dockstore-cli release 1.8.5 and then releases named the same as dockstore/dockstore for a couple of releases until most people who have an old dockstore-cli jar have upgraded. Then eventually we can figure out how to have the UI determine what the latest release is for dockstore-cli and point quickstart there to get the dockstore cli script. 

- Then we remove the dockstore-cli script from docstore/dockstore and provide some kind of message/documentation telling them to download the script from dockstore-cli repo if the user cannot upgrade.

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -104,6 +104,7 @@
     public static final AtomicBoolean INFO = new AtomicBoolean(false);
     public static final AtomicBoolean SCRIPT = new AtomicBoolean(false);
     public static final String DEPRECATED_PORT_MESSAGE = "Dockstore webservice has deprecated port 8443 and may remove it without warning. Please use 'https://dockstore.org/api' via the Dockstore config file (\"~/.dockstore/config\" by default) instead.";
+    public static final String DOCKSTORE_CLI_REPO_URL = "https://api.github.com/repos/dockstore/dockstore-cli/releases";
```
coverbeck, line 107: Should probably be private, but this is much better than the duplication that existed before.

denis-yuen:
> * Make a dockstore-cli release 1.8.5 and then releases named the same as dockstore/dockstore for a couple of releases until most people who have an old dockstore-cli jar have upgraded. Then eventually we can figure out how to have the UI determine what the latest release is for dockstore-cli and point quickstart there to get the dockstore cli script.
> * Then we remove the dockstore-cli script from docstore/dockstore and provide some kind of message/documentation telling them to download the script from dockstore-cli repo if the user cannot upgrade.

I think this is the tricky bit, we're already considering 1.8.6 so this redirect script would need to be brought forward for releases in the immediate future. Which is reasonable, but need to watch out.

denis-yuen:
> This looks good to me. It's the overall flow that I'm still a little fuzzy on -- the interaction with the script file.

So the script file is in a few places
* attached releases of the webservice
* installed in people's client machines

When the script file runs without a jar around, it downloads it. 
I don't recall if it actually downloads a new script to replace the old one (One second thought, I don't think it does).

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
wshands: made it private

wshands:
@denis-yuen @coverbeck What's the best way to move forward on this?

denis-yuen:
Could be a short storytime discussion to layout the issue and brainstorm a better solution.

wshands:
Story time bug explanation and fix proposal doc https://docs.google.com/document/d/1mASD-W8MOPxgauRCNhhz3TD91eLBzeN9q1EQihsU0-w/edit?usp=sharing

wshands:
I'd like to move forward with the fix outlined at story time. Is this acceptable?

PR #28, opened 2020-06-03
Bump webservice version

wshands:
Bump webservice version to latest


wshands:
not needed so closing

PR #31, opened 2020-06-16
don't replace backslashes

wshands:
Addresses:
dockstore/dockstore#3562
https://ucsc-cgl.atlassian.net/browse/DOCK-1416


wshands:
Ran a WDL workflow that utilized the GCS plugin and it ran successfully.

PR #39, opened 2020-08-31
Utilize the dockstore project BOM in the CLI project

wshands:
Version information for dependencies that are shared with the dockstore project are now stored in the dockstore project bom-internal pom.xml (BOM)

Addresses:
https://ucsc-cgl.atlassian.net/browse/SEAB-1629
https://ucsc-cgl.atlassian.net/browse/SEAB-1811

Includes fixes by Richard Hansen to address test failures; addresses:
https://ucsc-cgl.atlassian.net/browse/DOCK-1571


pom.xml
```diff
@@ -26,38 +26,17 @@
...
-        <mockito.version>2.28.2</mockito.version>
-        <cwlavro.version>2.0.4.4</cwlavro.version>
-        <okhttp.version>3.14.2</okhttp.version>
-        <slf4j.version>1.7.26</slf4j.version>
-        <swagger-ui.version>3.20.5</swagger-ui.version>
-        <maven-surefire.version>2.21.0</maven-surefire.version>
         <maven-failsafe.version>2.21.0</maven-failsafe.version>
-        <httpcomponents.version>4.5.10</httpcomponents.version>
-        <httpcore.version>4.4.12</httpcore.version>
-        <jetty-servlet.version>9.4.18.v20190429</jetty-servlet.version>
```
garyluu, line 54: Question for all, how come there's no warning like the web service? Also, shouldn't the CLI use a version that does not have a warning? https://github.com/dockstore/dockstore/network/alert/bom-internal/pom.xml/org.eclipse.jetty:jetty-server/open

denis-yuen: I don't know why the (dependabot?) warnings would be different.
I agree that we should use a version that solves the CVE.
(might need to be more than one PR)

garyluu: Looks like we don't have dependabot set up for this repo? that or I don't have permissions to see it

denis-yuen: FYI, added garyluu to permissions for dependabot

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -352,10 +352,6 @@ public void cwlVersion11() {
             .filter(version -> "master".equalsIgnoreCase(version.getName())).findFirst();
         assertTrue(optionalWorkflowVersion.isPresent());
         WorkflowVersion workflowVersion = optionalWorkflowVersion.get();
-        List<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
```
coverbeck, line 355: This may still be in progress, but there's a new endpoint that needs to be called where you need tol validate the sourcefiles.

coverbeck (changes requested):

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
wshands: Richard Hansen has fixed this

Richard-Hansen (approved):

dockstore-cli-integration-testing/pom.xml
```diff
@@ -56,6 +56,14 @@
             <groupId>io.dockstore</groupId>
             <artifactId>dockstore-webservice</artifactId>
             <version>${dockstore-core.version}</version>
+            <exclusions>
+                <exclusion>
+                    <!-- Solves the failure caused by two liquibase.build.properties files
+                     being included in the final JAR   -->
```
coverbeck, line 62: The file is in two separate JARs, not in the same JAR twice. And "the final JAR" may not be clear.

Suggest something like: "Liquibase is already bundled in the dockstore-webservice JAR".

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -470,21 +469,11 @@ private DockstoreTool getContainer() {
...
-        tag.setSourceFiles(list);
-        SourceFile fileDockerFile = new SourceFile();
-        fileDockerFile.setContent("dockerstuff");
-        fileDockerFile.setType(SourceFile.TypeEnum.DOCKERFILE);
-        fileDockerFile.setPath("/Dockerfile");
-        fileDockerFile.setAbsolutePath("/Dockerfile");
-        tag.getSourceFiles().add(fileDockerFile);
+        List<String> list = new ArrayList<>();
+        list.add("/Dockstore.cwl");
+        tag.setVerifiedSources(list);
```
coverbeck, line 474: This was adding to sourceFiles before, now it's adding to verifiedSources. Is this on purpose?

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralRegressionIT.java
```diff
@@ -112,21 +111,10 @@ private DockstoreTool getContainer() {
...
-        List<SourceFile> list = new ArrayList<>();
-        list.add(fileCWL);
-        tag.setSourceFiles(list);
-        SourceFile fileDockerFile = new SourceFile();
-        fileDockerFile.setContent("dockerstuff");
-        fileDockerFile.setType(SourceFile.TypeEnum.DOCKERFILE);
-        fileDockerFile.setPath("/Dockerfile");
-        fileDockerFile.setAbsolutePath("/Dockerfile");
-        tag.getSourceFiles().add(fileDockerFile);
+        List<String> list = new ArrayList<>();
```
coverbeck, line 114: Same as previous comment.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -753,7 +753,9 @@ public void handleInfo(String entryPath) {
                 errorMessage("This container is not published.", Client.COMMAND_ERROR);
             } else {
 
-                Date lastBuild = container.getLastBuild();
+                long lastBuildLong = container.getLastBuild();
+                Date lastBuild = new Date(lastBuildLong);
+
                 Date dateUploaded = null;
                 if (lastBuild != null) {
```
coverbeck, line 760: I thought I saw a warning about this from Codacy -- lastBuild can't be null.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -753,7 +753,9 @@ public void handleInfo(String entryPath) {
                 errorMessage("This container is not published.", Client.COMMAND_ERROR);
             } else {
 
-                Date lastBuild = container.getLastBuild();
+                long lastBuildLong = container.getLastBuild();
```
coverbeck, line 756: Suggest final

wshands: fixed

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
wshands: fixed by Richard Hansen

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralRegressionIT.java
wshands: fixed by Richard Hansen

dockstore-cli-integration-testing/pom.xml
wshands: Updated comment

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
wshands: I did not see this in Codacy; why couldn't lastBuild in fact be null? Comment at dockstore-webservice/src/main/resources/swagger.yaml:6666 says it can?

garyluu (commented):
There's a few `.toUpperCase()` .  It's not something I expect is needed.

For the descriptor language changes, one option would've just been to extend DescriptorLanguage and add back the missing function just for CLI purposes.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
denis-yuen: I think the idea is that even if `getLastBuild()` comes back from our API as null (not that Codacy understands our swagger probably) 
`Date lastBuild` won't be null because `new Date(null)` won't result in null (probably either exception or an epoch date of 1970ish)

coverbeck: 1. Sorry it was [lgtm](https://lgtm.com/projects/g/dockstore/dockstore-cli/rev/pr-1efc905f899a6835780c6ce185897bc84016c7b3), not Codacy -- I had had a vague recollection of having seen some email mentioning it.
2. `container.getLastBuild()` returns a Long, not a long, so on line 756 you should make it a Long. And because it's Long, you should do a null check there.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -753,7 +753,9 @@ public void handleInfo(String entryPath) {
                 errorMessage("This container is not published.", Client.COMMAND_ERROR);
             } else {
 
-                Date lastBuild = container.getLastBuild();
+                final long lastBuildLong = container.getLastBuild();
```
coverbeck, line 756: See comment below; this should be Long not long, and you should do a null check.

coverbeck (changes requested):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
wshands: fixed

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
wshands: Richard Hansen fixed it

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -753,9 +753,10 @@ public void handleInfo(String entryPath) {
                 errorMessage("This container is not published.", Client.COMMAND_ERROR);
             } else {
 
-                Date lastBuild = container.getLastBuild();
+                final Long lastBuildLong = container.getLastBuild();
                 Date dateUploaded = null;
-                if (lastBuild != null) {
+                if (lastBuildLong != null) {
+                    final Date lastBuild = new Date(lastBuildLong);
```
coverbeck, line 759: You could just do `dateUploaded = new Date(lastBuildLong)`, instead of creating an intermediate Date, but this is following the existing pattern, and maybe there's subtlety that I'm missing.

coverbeck (approved):

PR #40, opened 2020-09-04
Feature/inventory script

denis-yuen:
Working on https://ucsc-cgl.atlassian.net/browse/SEAB-1543

* prototype includes hook up for S3, EC2, and RDS
* will address other information in specific follow-up tickets


coverbeck (approved):

david4096 (commented):
Really excited to see you developing on these APIs! Both Charles and I have been meaning to I think and haven't got around to it.

It's not clear to me whether this should be in this repo though, won't this get included in future releases? Should this go to dockstore-deploy?

denis-yuen:
> It's not clear to me whether this should be in this repo though, won't this get included in future releases? Should this go to dockstore-deploy?

Was piggybacking on the code-style, findbugs, etc. in this repo
Right now this is a separate module disjoint from dockstore-cli (i.e. users won't download it as a new release of dockstore-cli)

dockstore-deploy could work too

coverbeck:
> dockstore-deploy could work too

Except for the CircleCI cost issue of private repos! :)

I was wondering about it being in this repo too, but seems OK since it's not bundled. Maybe we should eventually have a "utilities" repo, if we end up having more stuff like this.

denis-yuen:
@david4096 thoughts? can move if you feel strongly but need to figure out how to import/export/share plugin management configuration

david4096 (approved):
Can definitely stay here. Thank you for explaining the packaging concern.

PR #41, opened 2020-09-08
Feature/1423/improve refresh

Richard-Hansen:

- [Complete] Check that you pass the basic style checks and unit tests by running `mvn clean install` 

Ticket: https://ucsc-cgl.atlassian.net/browse/DOCK-1423

Functionality: Allows the CLI to discover workflows from repos that the user is added to while refreshing all workflows.

1-line change to the WorkflowClient of the CLI, added an integration test to match, which required a minor change to TestingPostgres.java to handle DELETE queries. 

I want to make sure the function 'runDeleteStatement' that was added to the TestingPostgres class does not skirt any quality or safety protocols.

### DIFF
--- a/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -70,6 +70,32 @@ public class GeneralWorkflowIT extends BaseIT {
         CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
     }
 
+    @Test
+    public void refreshAll() {
+        // refresh all to get a baseline number of workflows (we are testing the cli implementation of this)
+        refreshByOrganizationReplacement(USER_2_USERNAME);
+
+        // get userid
+        long userid = testingPostgres.runSelectStatement(String.format("SELECT id FROM user_profile WHERE username='%s';", USER_2_USERNAME), long.class);
+
+        // initial count of workflows associated with this user
+        int val1 = testingPostgres.runSelectStatement(String.format("SELECT COUNT(*) FROM user_entry WHERE userid = %d;", userid), int.class);
+
+        // Delete all user_entry rows associated with this user
+        testingPostgres.runDeleteStatement(String.format("DELETE FROM user_entry WHERE userid = %d", userid));
+
+        // ensure resulting count after delete is 0
+        int val2 = testingPostgres.runSelectStatement(String.format("SELECT COUNT(*) FROM user_entry WHERE userid = %d;", userid), int.class);
+        assertEquals("there should be 0 user entries for " + USER_2_USERNAME + ", there are " + val2, 0, val2);
+
+        // run CLI refresh command to refresh all workflows
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "refresh"});
+
+        // final count of workflows associated with this user
+        int val3 = testingPostgres.runSelectStatement(String.format("SELECT COUNT(*) FROM user_entry WHERE userid = %d;", userid), int.class);
+        assertEquals("Initially had " + val1 + " entries, resulted with " + val3 + " entries", val1, val3);
+    }
+
     /**
      * This test checks that refresh all workflows (with a mix of stub and full) and refresh individual.  It then tries to publish them
      */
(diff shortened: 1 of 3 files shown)


coverbeck:
> Please make sure that you've checked the following before submitting your pull request. Thanks!
> 
>     * [Complete] Check that you pass the basic style checks and unit tests by running `mvn clean install`

For future PRs, you should remove this (I used to leave it too)
 
> Ticket: https://ucsc-cgl.atlassian.net/browse/DOCK-1423

My personal reference is to use the GitHub issue if it's not a SEAB ticket (or you can do both). GitHub does seem neat linking when you link to a GitHub issue. There is short-hand to refer to issues; if in the dockstore/dockstore repo, just #<issue number>; if in a different repo, like here, dockstore/dockstore#<issue number>

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -70,6 +70,36 @@ public void resetDBBetweenTests() throws Exception {
...
+
+        // refresh all by organizations (We are testing the CLI implementation of this refresh functionality, we do the refresh here to give a baseline)
+        refreshByOrganizationReplacement(USER_2_USERNAME);
+
+        // initial count of workflows associated with this user using the standard refresh
+        final long val1 = testingPostgres.runSelectStatement(String.format("SELECT COUNT(*) FROM user_entry WHERE userid = %d;", userid), long.class);
+
+        // Delete all user_entry rows associated with repos this user is added to
+        final int deleteResult = testingPostgres.runDeleteStatement(String.format("DELETE FROM user_entry ue WHERE ue.userid = %d", userid));
+        assertEquals("After deletion, there should be 0 rows remaining associated with this user", deleteResult, val1);
```
coverbeck, line 89: I'm not sure you need line 85. Or, alternatively, maybe you don't need line 88.

If I'm reading this right, You've already deleted all the user entries on line 79. So I think you just want to assert that there are 0 entries associated with the user.

BTW, we're not consistent on this, but the second parameter in this case is the expected value, but you have it as the actual value. While it will fail either way if they are off, the message that is printed out makes it good to have the params in the correct order.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -70,6 +70,36 @@ public void resetDBBetweenTests() throws Exception {
...
+
+        // ensure resulting count after delete is less than the original count, otherwise we may not know if the final test is truly successful
+        final long val2 = testingPostgres.runSelectStatement(String.format("SELECT COUNT(*) FROM user_entry WHERE userid = %d;", userid), long.class);
+        assertTrue(String.format("There should be fewer entries post-deletion (pre-deletion, the user must be associated with > 0 workflows) pre: %d, post: %d", val1, val2), val1 > val2);
+
+        // run CLI refresh command to refresh all workflows
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "refresh"});
+
+        // final count of workflows associated with this user
+        final long val3 = testingPostgres.runSelectStatement(String.format("SELECT COUNT(*) FROM user_entry WHERE userid = %d;", userid), long.class);
```
coverbeck, line 99: Because you control the data, and know exactly what you should get, you should assert the expected values. It sort of goes against the grain of writing generic code, but you should do it this way in tests.

For example, if you expect there to be 3 workflows after this (don't know if that's the case, just making up a number), then the end-result should be you are testing on line 100 that 3 workflows are expected.

Also, may not be applicable depending on the changes you make based on my above comment, but I feel it's better to have a well named variable rather than `val3`. You can even do away with comments in many cases.

```
final long userWorkflowCount = ...
```

Then when I look at that variable's usage on line 100, I can tell what it is, rather than having go back up a line and see what `val3` is. It's a small thing, especially in this context, where the variable is on two consecutive lines, but in general, it's easier to read code without having to back and forth.


+1 on making all your variables final.

coverbeck (changes requested):

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
Richard-Hansen: I've updated the flow of the test case to be more direct. Workflows are initially registered, then all entries associated with the test user are deleted, refreshed, and checked for correctness.

Incorrect orderings of expected/actual parameters has been fixed.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
Richard-Hansen: Updated the test to now directly checks for the expected 40 linked workflows to the test user. Variable names have been re-written to be more self-documenting.

coverbeck (approved):

garyluu (approved):

PR #43, opened 2020-09-11
633/workflow entryname

Richard-Hansen:
Github Issue: dockstore/dockstore#2309

The user can specify a custom name for a workflow with the '--entryname' parameter when publishing the workflow from the CLI. This will create a separate workflow entry from the default registered workflow.

### DIFF
--- a/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -1077,10 +1077,10 @@ public class GeneralWorkflowIT extends BaseIT {
     }
 
     /**
-    * Tests publishing/unpublishing workflows with the --entryname parameter
+    * Tests publishing/unpublishing workflows with the --newEntryName parameter
     */
     @Test
-    public void testPublishWithEntryName() {
+    public void testPublishWithNewEntryName() {
         // register workflow
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "manual_publish", "--repository",
             "parameter_test_workflow", "--organization", "DockstoreTestUser2", "--git-version-control", "github", "--script"});
@@ -1094,7 +1094,7 @@ public class GeneralWorkflowIT extends BaseIT {
         // publish workflow with name 'test_entryname'
         Client.main(
             new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish",
-                "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", "--entryname", "test_entryname", "--script"});
+                "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", "--newEntryName", "test_entryname", "--script"});
 
         // verify there are 2 workflows associated with the user
         final long countTotalPublishedWorkflows = testingPostgres
@@ -1111,7 +1111,7 @@ public class GeneralWorkflowIT extends BaseIT {
         // Try unpublishing with both --unpub and --entryname specified, should fail
         systemExit.expectSystemExitWithStatus(Client.COMMAND_ERROR);
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish", "--unpub",
-            "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", "--entryname", "test_entryname", "--script"});
+            "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", "--newEntryName", "test_entryname", "--script"});
 
         // unpublish workflow with name 'test_entryname'
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish", "--unpub",
--- a/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -1081,6 +1081,9 @@ public class GeneralWorkflowIT extends BaseIT {
     */
     @Test
     public void testPublishWithNewEntryName() {
+
+        final String publishNameParameter = "--newEntryName";
+
         // register workflow
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "manual_publish", "--repository",
             "parameter_test_workflow", "--organization", "DockstoreTestUser2", "--git-version-control", "github", "--script"});
@@ -1094,7 +1097,15 @@ public class GeneralWorkflowIT extends BaseIT {
         // publish workflow with name 'test_entryname'
         Client.main(
             new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish",
-                "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", "--newEntryName", "test_entryname", "--script"});
+                "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", publishNameParameter, "test_entryname", "--script"});
+
+        // publish workflow with name 'test_entryname' a second time, shouldn't work
+        systemOutRule.clearLog();
+        Client.main(
+            new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish",
+                "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", publishNameParameter, "test_entryname", "--script"});
+        assertTrue("Attempting to publish a registered workflow should notify the user",
+            systemOutRule.getLog().contains("The following workflow is already registered: github.com/DockstoreTestUser2/parameter_test_workflow"));
 
         // verify there are 2 workflows associated with the user
         final long countTotalPublishedWorkflows = testingPostgres
@@ -1111,7 +1122,7 @@ public class GeneralWorkflowIT extends BaseIT {
         // Try unpublishing with both --unpub and --entryname specified, should fail
         systemExit.expectSystemExitWithStatus(Client.COMMAND_ERROR);
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish", "--unpub",
-            "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", "--newEntryName", "test_entryname", "--script"});
+            "--entry", "github.com/DockstoreTestUser2/parameter_test_workflow", publishNameParameter, "test_entryname", "--script"});
 
         // unpublish workflow with name 'test_entryname'
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish", "--unpub",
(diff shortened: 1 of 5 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -663,9 +664,29 @@ protected void handlePublishUnpublish(String entryPath, String newName, boolean
                     publish(true, entryPath);
                 }
             } else {
-                //for workflows method currently doesn't work with --entryname flag
-                errorMessage("Parameter '--entryname' not valid for workflows. See `workflow publish --help` for more information.",
-                    CLIENT_ERROR);
+                try {
```
coverbeck, line 667: GitHub doesn't let me comment on lines you haven't changed, but I'm reading the whole method to better understand what is going on, and noticed this: on line 650/651 the error message is saying it can't publish/unpublish `newName`, but it doesn't even use `newName` on the previous line. I think you want that to be `entryPath`.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -663,9 +664,29 @@ protected void handlePublishUnpublish(String entryPath, String newName, boolean
...
+                    final Workflow workflow = workflowsApi.getWorkflowByPath(entryPath, null, false);
```
coverbeck, line 668: This method is already invoked on line 648. Can't you just use `existingWorkflow`? Note, this is a network request, a relatively expensive call, so you don't want to duplicate this call if unnecessary.

Compiler may be possibly be giving you a variable undefined error, if so, but you should fix the code.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -663,9 +664,29 @@ protected void handlePublishUnpublish(String entryPath, String newName, boolean
...
+                            workflow.getOrganization() + "/" + workflow.getRepository(),
+                            workflow.getWorkflowPath(),
+                            newName,
+                            workflow.getDescriptorType().toString(),
+                            workflow.getDefaultTestParameterFilePath()
+                    );
+
+                    if (newWorkflow != null) {
+                        out("Successfully registered " + entryPath + "/" + newName);
+                        workflowsApi.refresh(newWorkflow.getId());
```
coverbeck, line 682: Is this line necessary? I think it is, but just double-checking.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
coverbeck, line 667: This was also like this, but for the block starting on line 652/653, I feel that we should have a message saying `newName` is not valid for unpublish.

coverbeck (changes requested):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: The current implementation doesn't handle publishing/unpublishing workflows with '--entryname' gracefully if there is an error. For example, if I try to publish a workflow with a custom name that is already published, the function will throw an ApiException instead of printing a clean "This workflow is already published" message.

I'll fix this in my next commit, but would like to clarify expected functionality:
`dockstore workflow publish --unpub --entry workflowpath --entryname newname` 
vs 
`dockstore workflow publish --unpub --entry workflowpath/entryname`

Currently, in the first case, the entry name is ignored and the command is interpreted as: 
`dockstore workflow publish --unpub --entry workflowpath`.
 Would it make sense to have the two cases listed above be logically equivalent in execution? Or should the function prevent use of --unpub with --entryname?

coverbeck: I will defer to @garyluu and @denis-yuen , but here are my thoughts:

1. If we're going to support `--entryname` we should add it to the [help message](https://github.com/dockstore/dockstore-cli/blob/5de67a497465612ddc693ffc694e916eea018af2/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L689). I didn't realize until reading the issue again that this an enhancement.
2. Depending on how you want to approach it, either that help message should say that the argument is ignored for unpublish; or, the help message says it's only applicable to publish and if we detect it here for unpublish, we throw an error.

In either case, I think this renders your question moot; --entryname should either be ignored or cause an error if used with unpublish.

In hindsight, I wished we had called it something like `newEntryName`.

garyluu: I'm personally not a fan of flags that conflict with other flags when it can be helped. I'd rather there be an unpublish command. However, in order to keep things similar to tools, I suppose that shouldn't be done.

Anyways, given the current situation, I think I agree with Charles. What does tools do right now?

Richard-Hansen: Tools functions in a [similar way](https://github.com/dockstore/dockstore-cli/blob/5de67a497465612ddc693ffc694e916eea018af2/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java#L213) to workflows. The `--entryname` parameter is ignored for unpublish, regardless of if it is set or not. Both could use slightly more descriptive error messages for when publish/unpublish fails.

I can modify workflows to better handle `--entryname` with unpublish and provide more descriptive errors. Should i do the same for tools/create a similar ticket for tools? Furthermore, I think @coverbeck's idea to change `--entryname` to something such as `--newEntryName` would be beneficial.

coverbeck: The problem is, even though it's extremely unlikely, is that there could be some script somewhere using `--entryname`, and changing the parameter name would break that script. Some ideas:

1. Only make it `newEntryName` for workflows, since it's a new feature there. The downside is that it's asymmetrical with tools though.
2. Make `--newEntryName` an additional option for tools. Continue to support `--entryname`, but undocumented, i.e., the help message for tools only lists `--newEntryName`.
3. It's unlikely anybody is using this feature, just change it on tools.

I suggest the first option for now, and wait for Denis' opinion.

garyluu: I think if you have time to make all the changes (for both tools and workflows) mentioned by Charles then go ahead. Otherwise, I think you should first match the functionality of tools and workflows.   And then create tickets to simultaneously make improvements to both.

Richard-Hansen: I'll update both tools and workflows to have the same handling of the `--entryname` parameter, to ensure they are logically symmetric. I'll move forward from there after getting Denis' opinion.

Richard-Hansen: Updated. Now reports that the error occurred when trying to locate the registered workflow at `entryPath`.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: Updated. No longer makes the additional unnecessary API call.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: Ideally, manualRegister will not return a null value, and throw the expected API exception. In the theoretical case it doesn't, this is extra verbose.

coverbeck (approved):
LGTM to me, but pending decision on entryname when Denis gets back.

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
denis-yuen: This looks like its going into develop ... I'm ok with 2. actually (rename to better parameter name but allow old name for backwards compatibility). Also ok with a series of PRs for the same reason (I've be more nervous about a partial incomplete fix for 1.9.x)

Richard-Hansen: Ok, I'll update `--entryName` to `--newEntryName`, and modify the help messages to specify `--newEntryName` but still keep support for `--entryName`.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -449,9 +450,21 @@ public void publish(List<String> args) {
...
         } else {
             String first = reqVal(args, "--entry");
-            String entryname = optVal(args, "--entryname", null);
             final boolean unpublishRequest = args.contains("--unpub");
-            handlePublishUnpublish(first, entryname, unpublishRequest);
+
+            // --newEntryName is the desired parameter flag, but maintaining backwards compatibility for --entryname
+            String newEntryName = optVal(args, "--newEntryName", null);
+            if (newEntryName == null) {
+                newEntryName = optVal(args, "--entryname", null);
```
coverbeck, line 458: 1. Perhaps a warning that `--entryName` is deprecated? 
2. Also, checking for it here, you're going to be supporting --entryName for workflows. I thought the idea was tool would support --entryName and --newEntryName, because tool already has --entryName, but since it's being added to workflow for the first time, we don't have to worry about backwards compatibility.

I don't feel too strongly about either of these.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1075,4 +1075,51 @@ public void testGenerateDOIFrozenVersion() throws ApiException {
         }
 
     }
+
+    /**
+    * Tests publishing/unpublishing workflows with the --newEntryName parameter
+    */
+    @Test
+    public void testPublishWithNewEntryName() {
```
coverbeck, line 1083: Would be good to have tests that verify both --newEntryName and --entryName work (all these tests are for workflow).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -210,28 +211,53 @@ protected void handleSearch(String pattern) {
...
+        assert (!(unpublishRequest && newName != null));
+
+        try {
+            existingTool = containersApi.getContainerByToolPath(entryPath, null);
+            isPublished = existingTool.isIsPublished();
+        } catch (ApiException ex) {
+            exceptionMessage(ex, "Unable to " + (unpublishRequest ? "unpublish " : "publish ") + entryPath, Client.API_ERROR);
+        }
+
+        if (existingTool == null) {
```
coverbeck, line 227: I don't think this if test is necessary. If line 221 fails to find a tool, it will throw an exception, and line 224 will exit. And if it didn't throw an exception, you'd get an NPE on line 222.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -641,35 +642,75 @@ protected void refreshTargetEntry(String path) {
...
+
         try {
             existingWorkflow = workflowsApi.getWorkflowByPath(entryPath, null, false);
             isPublished = existingWorkflow.isIsPublished();
         } catch (ApiException ex) {
-            exceptionMessage(ex, "Unable to publish/unpublish " + newName, Client.API_ERROR);
+            exceptionMessage(ex, "Unable to " + (unpublishRequest ? "unpublish " : "publish ") + entryPath, Client.API_ERROR);
+        }
+
+        if (existingWorkflow == null) {
```
coverbeck, line 658: Same as other comment; I don't think you need this block.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -641,35 +642,75 @@ protected void refreshTargetEntry(String path) {
...
+                            existingWorkflow.getOrganization() + "/" + existingWorkflow.getRepository(),
+                            existingWorkflow.getWorkflowPath(),
+                            newName,
+                            existingWorkflow.getDescriptorType().toString(),
+                            existingWorkflow.getDefaultTestParameterFilePath()
+                    );
+
+                    final String completeEntryPath = entryPath + "/" + newName;
+
+                    if (newWorkflow != null) {
```
coverbeck, line 689: Won't newWorkflow always be set? If line 678 failed, won't it throw an exception?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: 1. Adding a deprecation warning
2. My initial thinking was that it would seem more consistent to have `--entryname` compatible, but deprecated, for both tools and workflows. Since this is currently (happy to change it) being handled within the AbstractEntryClient, I opted to avoid specifying tools/workflows specifically. Thoughts?

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
Richard-Hansen: Adding tests for both `--entryname` and `--newEntryName` for tools and workflows.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
Richard-Hansen: I see what you're saying, thanks for the catch! Removing the unnecessary lines.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: Agreed. Same as above.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: Ideally, yes. I added this because if, in some strange case, the `manualRegister()` API call had null as its return value, that would not throw an exception from the register function itself (We would hit a null pointer exception 2 lines later, so it wouldn't make it very far).

Either way, it is redundant, I will remove it from both tools and workflows.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -449,9 +451,22 @@ public void publish(List<String> args) {
             publishHelp();
         } else {
             String first = reqVal(args, "--entry");
-            String entryname = optVal(args, "--entryname", null);
             final boolean unpublishRequest = args.contains("--unpub");
-            handlePublishUnpublish(first, entryname, unpublishRequest);
+
+            // --newEntryName is the desired parameter flag, but maintaining backwards compatibility for --entryname
+            String newEntryName = optVal(args, "--newEntryName", null);
```
coverbeck, line 457: Unless we already do this elsewhere, I don't think we add mixed case parameter names to the command-line.

coverbeck (commented):
Looks good, but I'd missed the mixed case parameter the first time through. Even we lower-case the args in the code (do we?) before testing, we should keep the same patterns throughout.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: Looks like dash-delimited is the most common, `word-word-word`. I'll switch to that pattern. Thanks!

coverbeck:
> This pull request **introduces 2 alerts** when merging [85ee68e](https://github.com/dockstore/dockstore-cli/commit/85ee68e1a3818df1e5cf5b1f94f5f0847d7d3b8e) into [4a9474a](https://github.com/dockstore/dockstore-cli/commit/4a9474a7eae4e5d38f317a102e57593510d002af) - [view on LGTM.com](https://lgtm.com/projects/g/dockstore/dockstore-cli/rev/pr-0c5438b436f20169e385eb30968ba3cdcc6e10d3)
> 
> **new alerts:**
> 
>     * 2 for Dereferenced variable may be null

Sorry, looks like that was my fault. :(

denis-yuen (approved):

garyluu (approved):

PR #45, opened 2020-09-25
Feature/3664/validate descriptor checksum

Richard-Hansen:
Github issue: dockstore/dockstore#3664

When running the launch command for tools and workflows, primary and secondary descriptors are validated using TRS endpoints. The validation can be ignore by supplying the `--ignore-checksums` flag to the launch command.

Some other functions were modified to ensure that the version that is downloaded is the version that is requested via the TRS endpoint. See [workflows](https://github.com/dockstore/dockstore-cli/blob/1b6708fe7de858cb7b578c20d7f0c3b5b50ec75e/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L355) and [tools](https://github.com/dockstore/dockstore-cli/blob/1b6708fe7de858cb7b578c20d7f0c3b5b50ec75e/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java#L684).

A few tests were modified to either ignore checksum validation, or to not directly call the launchPipeline function (this circumvents the instantiation of the GA4GHV20 API object).


dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -1498,6 +1498,39 @@ public void testRefreshingUserMetadata() {
...
+        // refresh the tool
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "tool", "refresh", "--entry",
+            "quay.io/dockstoretestuser/test_input_json", "--script" });
+
+        // launch the tool
+        systemOutRule.clearLog();
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "tool", "launch", "--entry",
+            "quay.io/dockstoretestuser/test_input_json", "--json", ResourceHelpers.resourceFilePath("tool_hello_world.json"), "--script" });
+        assertTrue("Output should indicate that checksums have been validated",
+            systemOutRule.getLog().contains("Validated checksums") && !systemOutRule.getLog().contains("Cannot validate the checksum of the locally downloaded descriptor."));
```
denis-yuen, line 1516: Use constant string shared with implementation so that future developers can tell that this test is associated with the implementation below and to avoid instant breakage

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1096,7 +1096,47 @@ public void testGenerateDOIFrozenVersion() throws ApiException {
...
+            new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "manual_publish", "--repository",
+                "md5sum-checker", "--organization", "DockstoreTestUser2", "--git-version-control", "github",
+                "--workflow-path", "/checker-workflow-wrapping-tool.cwl", "--descriptor-type", "cwl", "--workflow-name", "checksumTester", "--script" });
+
+        // ensure checksum validation is acknowledged, and no null checksums were discovered
+        systemOutRule.clearLog();
+        Client.main(new String[] {"--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "launch", "--entry",
+            "github.com/DockstoreTestUser2/md5sum-checker/checksumTester", "--json", ResourceHelpers.resourceFilePath("md5sum_cwl.json"), "--script" });
+        assertTrue("Output should indicate that checksums have been validated",
+            systemOutRule.getLog().contains("Validated checksums") && !systemOutRule.getLog().contains("Cannot validate the checksum of the locally downloaded descriptor."));
```
denis-yuen, line 1120: Ditto for these tests

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -1498,6 +1498,39 @@ public void testRefreshingUserMetadata() {
...
+
+        systemOutRule.clearLog();
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "tool", "launch", "--entry",
+            "quay.io/dockstoretestuser/test_input_json", "--json", ResourceHelpers.resourceFilePath("tool_hello_world.json"), "--script" });
+        assertTrue("Output should indicate that checksums have been validated",
+            systemOutRule.getLog().contains("Validated checksums") && systemOutRule.getLog().contains("Cannot validate the checksum of the locally downloaded descriptor."));
+
+        testingPostgres.runUpdateStatement("UPDATE sourcefile SET checksums = 'SHA-1:VeryFakeChecksum' WHERE path='/Dockstore.cwl';");
+
+        systemExit.expectSystemExitWithStatus(Client.API_ERROR);
```
denis-yuen, line 1529: Wonder if you can test the exit message (Cannot validate the checksum ...) with https://stefanbirkner.github.io/system-rules/apidocs/org/junit/contrib/java/lang/system/ExpectedSystemExit.html#checkAssertionAfterwards-org.junit.contrib.java.lang.system.Assertion-

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -386,6 +381,69 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+        final Workflow workflow = getDockstoreWorkflowByPath(parts[0]);
+
+        // ensure workflow has version
+        Optional<WorkflowVersion> first = workflow.getWorkflowVersions().stream().filter(foo -> foo.getName().equalsIgnoreCase(versionID))
+            .findFirst();
+
+        // if no master is present (for example, for hosted workflows), fail over to the latest descriptor
+        if (first.isEmpty()) {
+            first = workflow.getWorkflowVersions().stream().max(Comparator.comparing(WorkflowVersion::getLastModified));
+            first.ifPresent(workflowVersion -> System.out.println(
```
coverbeck, line 428: Don't use System.out.println, use out

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -386,6 +381,69 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+                "Could not locate workflow with version '" + versionID + "'. Using last modified version '" + workflowVersion.getName()
+                    + "' instead."));
+            return first.get().getName();
```
coverbeck, line 431: Will NPE if `first` is empty. And I think we do have edge cases of workflows with no versions.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -386,6 +381,69 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+        return versionID;
+    }
+
+    /**
+     * Returns if the provided tool path is for a published entry
+     * @param entryPath Tool path
+     */
+    @Override
+    public boolean isEntryPublished(String entryPath) {
+        Workflow workflow = getDockstoreWorkflowByPath(entryPath);
```
coverbeck, line 443: I don't want to over-optimize this, especially given all the individual checksum calls you'll need to make, and the actual running of the workflow proper, but fetching a workflow is a relatively expensive operation.

Seems like you may be calling it 3 times (I didn't step through it)? If you have to jump through hoops to restructure the code, don't do it, but if it's easy, it would be nicer if you could reuse the Workflow object.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
Richard-Hansen: [fixed](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java#L1505). Added final strings at start of test.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
Richard-Hansen: [Fixed](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java#L1110). Added final strings at start of tests.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
Richard-Hansen: Yes! Now, after the expected system exit, the expected output is [checked](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java#L1143). Used for Tool test as well.

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -792,7 +794,8 @@ public void setupClientEnvironment(List<String> args) {
         ApiClient defaultApiClient;
         defaultApiClient = Configuration.getDefaultApiClient();
         String cliVersion = getClientVersion();
-        defaultApiClient.setUserAgent("Dockstore-CLI/" + cliVersion + "/java");
+        final String userAgent = "Dockstore-CLI/" + cliVersion + "/java";
+        defaultApiClient.setUserAgent(userAgent);
 
         ApiKeyAuth bearer = (ApiKeyAuth)defaultApiClient.getAuthentication("BEARER");
```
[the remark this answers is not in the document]
Richard-Hansen: [Added OAuth token](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java#L815). TRS endpoints only work for published entries, so this change shouldn't impact the current implementation, but as we move more towards OpenAPI, the openapi.client.ApiClient class will be used more.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -714,6 +707,58 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+        // last resort use 'latest' tag
+        return versionID == null ? "latest" : versionID;
+    }
+
+    /**
+     * Returns if the provided tool path is for a published entry
+     * @param entryPath Tool path
+     */
+    @Override
+    public boolean isEntryPublished(String entryPath) {
```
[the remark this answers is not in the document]
Richard-Hansen: I haven't addressed this in the current implementation yet. 

Because TRS endpoints only work for published entries, what is the desired way to handle this? I can make an API call for the remote descriptor file (full-text version), and locally calculate the SHA-1 checksum? ([Instead of using the GA4GHV20 call](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L234))

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,76 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+
+                remoteDescriptorChecksum = remoteChecksumList.size() > 0 ? remoteChecksumList.get(0) : null;
+            } catch (io.dockstore.openapi.client.ApiException ex) {
+                // The TRS endpoint throw openapi ApiExceptions
+                exceptionMessage(ex, "unable to locate remote descriptor " + ga4ghv20Path, ENTRY_NOT_FOUND);
+            }
+
+            // Get local descriptor checksum
+            try {
+                final File localDescriptor = new File(localTemporaryDirectory + toolFile.getPath());
```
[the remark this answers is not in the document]
Richard-Hansen: [Fixed](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L263). According to documentation (and manual testing) if the child file path is an absolute path, it will be converted to a relative path, so there should be no issues.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,76 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+                exceptionMessage(ex, "unable to locate local descriptor at " + localTemporaryDirectory + toolFile.getPath(), IO_ERROR);
+            }
+
+            // compare checksums
+            if (remoteDescriptorChecksum == null) {
+                // no stored descriptor
+                err("Unable to locate checksum for descriptor " + toolFile.getPath() + ". Cannot validate the checksum of the locally downloaded descriptor. Refreshing the workflow will calculate the entry checksum.");
+            } else if (!remoteDescriptorChecksum.equals(localDescriptorChecksum)) {
+                // checksums do not match
+                errorMessage("Local checksum does not match remote checksum. Launch halted.", API_ERROR);
```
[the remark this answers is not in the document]
Richard-Hansen: [Added](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L278). Good call, thanks!

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: [Fixed](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L430).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: [Fixed](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L435).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,76 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+                exceptionMessage(ex, "unable to locate remote descriptor " + ga4ghv20Path, ENTRY_NOT_FOUND);
+            }
+
+            // Get local descriptor checksum
+            try {
+                final File localDescriptor = new File(localTemporaryDirectory + toolFile.getPath());
+                final String fileContents = FileUtils.readFileToString(localDescriptor, StandardCharsets.UTF_8);
+                final String fileChecksum = DigestUtils.sha1Hex(fileContents);
+                localDescriptorChecksum = (new Checksum()).checksum(fileChecksum);
+                localDescriptorChecksum.setType("sha1");
```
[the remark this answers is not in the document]
Richard-Hansen: I believe this is handled [here](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L252). The TRS endpoint returns a list of checksums, and we filter for all SHA-1 checksums.

So I should only be comparing the local descriptor with whatever is stored as a SHA-1 checksum. I can expand this to instead determine the types of checksums that are available (i.e. we can look at the checksums returned [here](https://github.com/dockstore/dockstore-cli/blob/7f6c85a4a186819af1abeaf2f795939c1fb693b4/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L249), and determine which checksum I should calculate from whatever is return from the API call?). Currently, even if checksums are returned, if none of them are SHA-1, the remoteDescriptorChecksum will be set to null.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: I'm happy to refactor. Would it be better to pass the entry object around, or add a variable to the abstract entry client to store the 'current' entry being requested (this could also allow for refactoring elsewhere.)

i.e. if you call `workflow launch --entry entryname --json ...`, I can fetch the workflow specified by `entryname` and store that in the WorkflowClient object. This also means that we could change how other various functions fetch workflows, and potential reduce calls across the CLI. I can also add this as another ticket. Thoughts?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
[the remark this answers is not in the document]
denis-yuen: One other option, extend TRS endpoint to accept auth to access unpublished workflows (might be a big task that flows into a different ticket)

denis-yuen (approved):

coverbeck (commented):
I know there's a lot of nit-picky stuff in my comments. Consider most of it optional. Except for the published / authorization part -- we need to make sure that's right.

Overall this is really good, the nit-picky stuff is just a teaching opportunity for future reference. :)

denis-yuen:
Thanks Charles, totally forgot that I suggested/was involved in setting up optional auth for TRS 
https://github.com/dockstore/dockstore/pull/1614

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -1498,6 +1498,50 @@ public void testRefreshingUserMetadata() {
         Assert.assertEquals("One user should have this info now, there are " + count, 1, count);
     }
 
+    @Test
+    public void launchToolChecksumValidation() {
+
+        // These match the print statements made by the validateDescriptorChecksums function
+        final String checksumsValidatedPrint = "Validated checksums";
```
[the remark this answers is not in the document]
denis-yuen: To add to this, even if the tests would catch a mis-match, its an annoying catch and it is time consuming.
Better to signal to future-us that the test and implemention are tied by using the same string constant

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,76 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+                    .filter(c -> c.getType().equals("sha1"))
+                    .collect(Collectors.toList());
+
+                remoteDescriptorChecksum = remoteChecksumList.size() > 0 ? remoteChecksumList.get(0) : null;
+            } catch (io.dockstore.openapi.client.ApiException ex) {
+                exceptionMessage(ex, "unable to locate remote descriptor " + ga4ghv20Path, ENTRY_NOT_FOUND);
+            }
+
+            // Get local descriptor checksum
+            try {
```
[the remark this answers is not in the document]
Richard-Hansen: [Added!](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L259) Good call.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,76 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+                // if the toolFile.getPath() is absolute, it is converted to a relative path by File the constructor
+                final File localDescriptor = new File(localTemporaryDirectory, toolFile.getPath());
+                final String fileContents = FileUtils.readFileToString(localDescriptor, StandardCharsets.UTF_8);
+                final String fileChecksum = DigestUtils.sha1Hex(fileContents);
+                localDescriptorChecksum = (new Checksum()).checksum(fileChecksum);
+                localDescriptorChecksum.setType("sha1");
```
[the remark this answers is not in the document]
Richard-Hansen: [Refactored](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L237).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,76 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+
+        // Get the entry version we are validating for
+        String versionID = abstractEntryClient.getVersionID(entryVal);
+
+        List<ToolFile> allDescriptors = abstractEntryClient.getAllToolDescriptors(type.toString(), path, versionID);
+
+        Ga4Ghv20Api ga4ghv20api = abstractEntryClient.getClient().getGa4Ghv20Api();
+
+        // All secondary files are located relative to the location of the primary descriptor.
+        final String localTemporaryDirectory = localPrimaryDescriptorFile.getParent() + "/";
```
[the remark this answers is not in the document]
Richard-Hansen: [Removed](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L236). Good call, especially once switching to getting the File object via File(dir, child).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -714,6 +707,60 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+     */
+    @Override
+    public List<ToolFile> getAllToolDescriptors(String type, String entryPath, String versionID) {
+
+        final Ga4Ghv20Api ga4ghv20api = this.client.getGa4Ghv20Api();
+
+        // get all the tool files and filter out anything not a descriptor
+        try {
+            return ga4ghv20api.toolsIdVersionsVersionIdTypeFilesGet(type, entryPath, versionID).stream()
+                .filter(toolFile -> toolFile.getFileType().equals(ToolFile.FileTypeEnum.SECONDARY_DESCRIPTOR) || toolFile.getFileType().equals(ToolFile.FileTypeEnum.PRIMARY_DESCRIPTOR))
```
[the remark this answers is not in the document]
Richard-Hansen: [Fixed](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java#L808). This is great, definitely something to apply in the future.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -386,6 +381,71 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+     * @param type workflow type, CWL, WDL, NFL ...
+     * @param entryPath Workflow path
+     * @param versionID version we are getting descriptors for
+     */
+    public List<ToolFile> getAllToolDescriptors(String type, String entryPath, String versionID) {
+
+        final Ga4Ghv20Api ga4ghv20api = this.client.getGa4Ghv20Api();
+
+        // workflows are identified within ga4gh endpoints when their prefix is '#workflow/' (tools do not require a special prefix)
+        final String ga4ghv20Path = "#workflow/" + entryPath;
```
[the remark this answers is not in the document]
Richard-Hansen: [Moved to here](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java#L802), implemented in AbstractEntryClient, and a [getTrsId()](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java#L792) function has been added and overridden for workflows.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,76 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+        // Validate each tool file associated with the entry (Primary and secondary descriptors)
+        for (ToolFile toolFile : allDescriptors) {
+            Checksum remoteDescriptorChecksum = null;
+            Checksum localDescriptorChecksum = null;
+
+            // Get remote descriptor checksum
+            try {
+                // The TRS endpoint only discovers published entries
+                final FileWrapper remoteDescriptor = ga4ghv20api.toolsIdVersionsVersionIdTypeDescriptorRelativePathGet(type.toString(), ga4ghv20Path, versionID, toolFile.getPath());
+                List<Checksum> remoteChecksumList = remoteDescriptor.getChecksum()
```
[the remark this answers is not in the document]
Richard-Hansen: I kept as-is for now. I think keeping both checksums as the same type is nice. Happy to modify anyways, if the other approach is more standard/preferred, or potentially better?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: This function has been removed. Refactoring further may be challenging, as the [descriptors downloaded](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L353) could potentially be a different version than the one specified in the entryPath. This causes the additional call to [this](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java#L455).

Richard-Hansen:
[This](https://github.com/dockstore/dockstore-cli/blob/508cfdb2362f8d01a989b61680a295ac326def5f/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java#L814) is how the user token is added for API calls. Not necessarily an issue, just different than for the swagger API client. Thought it was worth pointing out. Thanks for helping solve this, @coverbeck.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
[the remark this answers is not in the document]
coverbeck: > I kept as-is for now. I think keeping both checksums as the same type is nice. Happy to modify anyways, if the other approach is more standard/preferred, or potentially better?

My opinion on nulls:

They're sometimes unavoidable, e.g., when using 3rd party libraries. And in small blocks of code it may not matter that much. But in general nulls are evil and I prefer to avoid them at all costs. :) If you have a possible null,  you have to check that it's not null. More crucially, you have to remember to check for it. Failures to check for null are probably the most common type of runtime error that occur in Java.

Whereas with Optional, you know you have to check for it for being empty -- the compiler enforces it.

It's more crucial when you return a value from a method that could be null. If you're using it within a method, like here, it's less of an issue.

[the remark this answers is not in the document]
Richard-Hansen: I can always support avoiding NPEs!

Refactored [here](https://github.com/dockstore/dockstore-cli/blob/8e1a149956ea321769c5bc7340e41e081b5bee37/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L243) to use Optional\<Checksum\> for the remote descriptor. Also switched around the checksum comparison block to avoid additional conditional checks.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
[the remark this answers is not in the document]
coverbeck: I think you may have overlooked these comments.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -386,6 +379,43 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+
+        // if no master is present (for example, for hosted workflows), fail over to the latest descriptor
+        if (first.isEmpty()) {
+            first = workflow.getWorkflowVersions().stream().max(Comparator.comparing(WorkflowVersion::getLastModified));
+            first.ifPresent(workflowVersion -> out(
+                "Could not locate workflow with version '" + versionID + "'. Using last modified version '" + workflowVersion.getName()
+                    + "' instead."));
+        }
+
+        return first.isEmpty() ? versionID : first.get().getName();
```
coverbeck, line 416: I don't grok the entire flow, but you are potentially returning the version id of a non-existent version. Is that going to work further down the line?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -714,6 +705,25 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
+
+        String versionID = parts.length > 1 ? parts[1] : null;
+
+        if (versionID == null) {
+            final DockstoreTool container = getDockstoreTool(parts[0]);
+            versionID = container.getDefaultVersion();
+        }
+
+        // last resort use 'latest' tag
+        return versionID == null ? "latest" : versionID;
```
coverbeck, line 724: What if latest doesn't exist?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -770,6 +784,38 @@ protected void testParameter(List<String> args) {
...
+     */
+    public List<ToolFile> getAllToolDescriptors(String type, String entryPath, String versionID) {
+        final Ga4Ghv20Api ga4ghv20api = this.getClient().getGa4Ghv20Api();
+
+        // get all the tool files and filter out anything not a descriptor
+        try {
+            return ga4ghv20api.toolsIdVersionsVersionIdTypeFilesGet(type, entryPath, versionID).stream()
+                .filter(toolFile -> ToolFile.FileTypeEnum.SECONDARY_DESCRIPTOR.equals(toolFile.getFileType()) || ToolFile.FileTypeEnum.PRIMARY_DESCRIPTOR.equals(toolFile.getFileType()))
+                .collect(Collectors.toList());
+        } catch (NullPointerException ex) {
```
coverbeck, line 810: Have you seen an NPE happen? You need to catch if so, but just wondering why you need this.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
coverbeck: Looking at the code, I think it's OK; the GA4GH call will fail and you'll exit.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -199,6 +212,74 @@ public long launchPipeline(String entryVal, boolean localEntry, String yamlFile,
...
+                Checksum localDescriptorChecksum = null;
+                try {
+                    // if the toolFile.getPath() is absolute, it is converted to a relative path by File the constructor
+                    final File localDescriptor = new File(localTemporaryDirectory, toolFile.getPath());
+                    final String fileContents = FileUtils.readFileToString(localDescriptor, StandardCharsets.UTF_8);
+                    final String fileChecksum = DigestUtils.sha1Hex(fileContents);
+                    localDescriptorChecksum = (new Checksum()).checksum(fileChecksum);
+                    localDescriptorChecksum.setType(checksumFunction);
+                } catch (IOException ex) {
+                    exceptionMessage(ex, "unable to locate local descriptor at " + localTemporaryDirectory + "/" + toolFile.getPath(), IO_ERROR);
```
coverbeck, line 267: You already have `localDescriptor`, you're sort of reconstructing it here again.

Also I would make this and other error messages sentences, or at least start the phrase with a capital. You could argue it either way, but we should be consistent, and a quick grep shows most of our error messages start with capitals.

coverbeck (commented):
More nits; really close. :)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
Richard-Hansen: [Updated](https://github.com/dockstore/dockstore-cli/blob/2a2f2ea14f509729cf60b324bfd05beb55fac418/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java#L714). This now more closely resembles the implementation in the WorkflowClient class, while still conforming to the original standard from [here](https://github.com/dockstore/dockstore-cli/blob/e05578d5a95d3f57757939b89ee280fb53a717a1/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java#L680). Similar to that in WorkflowClient, If there is no latest/default version, we will default to the version that was built most recently.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: This was, when looking back, caused by an issue addressed in an earlier comment. By changing the order in which `ToolFile.FileTypeEnum.SECONDARY_DESCRIPTOR.equals(toolFile.getFileType())` is checked, the NPE is no longer present. Removed the catch.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
Richard-Hansen: [Fixed](https://github.com/dockstore/dockstore-cli/blob/2a2f2ea14f509729cf60b324bfd05beb55fac418/dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java#L261).

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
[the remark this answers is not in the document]
Richard-Hansen: I did! My fault, thanks for bringing them to my attention.

Modified where the messages are defined to [here](https://github.com/dockstore/dockstore-cli/blob/2a2f2ea14f509729cf60b324bfd05beb55fac418/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java#L126).

denis-yuen (approved):
When builds are passing

coverbeck (approved):
Pending passing tests.

wshands (approved):

garyluu (approved):

PR #46, opened 2020-10-07
Feature/3795/ignoreSourceFileFixes

Richard-Hansen:
Initially just CLI changes for dockstore/dockstore#3795, but covers other issues associated with using the newer version of the webservice (1.10.0-alpha.1) with the CLI.

### DIFF
--- a/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -289,10 +289,10 @@ public class GeneralWorkflowIT extends BaseIT {
                 "testname", "--workflow-path", "/Dockstore.wdl", "--descriptor-type", "wdl", "--script" });
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "version_tag", "--entry",
             SourceControl.GITHUB.toString() + "/DockstoreTestUser2/hello-dockstore-workflow/testname", "--name", "master",
-            "--workflow-path", "/Dockstore2.wdl", "--hidden", "true", "--script" });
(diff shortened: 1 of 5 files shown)


dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -289,10 +289,10 @@ public void testUpdateWorkflowVersion() {
                 "testname", "--workflow-path", "/Dockstore.wdl", "--descriptor-type", "wdl", "--script" });
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "version_tag", "--entry",
             SourceControl.GITHUB.toString() + "/DockstoreTestUser2/hello-dockstore-workflow/testname", "--name", "master",
-            "--workflow-path", "/Dockstore2.wdl", "--hidden", "true", "--script" });
+            "--workflow-path", "/Dockstore2.wdl", "--script" });
```
coverbeck, line 292: Why is this being removed?

coverbeck (approved):
FYI, you don't need approval to merge to the feature branch (although it doesn't hurt if you want input). It's going to reviewed all over again on the merge from that feature branch to develop.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
Richard-Hansen: Some tests were failing due to setting the default version of an entry to hidden. This is why [this](https://github.com/dockstore/dockstore-cli/pull/46/files#diff-6f92ba8b67e3825333d92f85481441caR106) was changed as well. 

Looks to have been implemented in dockstore/dockstore#3708

PR #47, opened 2020-10-13
Test remove exclusion of liquibase build properties

wshands:
Need to confirm that this exclusion is necessary for integration tests to pass


wshands:
Closing because it confirms the exclusion is necessary

PR #48, opened 2020-10-14
Feature/3591/update cwl tool

garyluu:
For dockstore/dockstore#3591

There were several unrelated breaking changes from before (hard refresh query parameter). I set some to null instead of true to mimic the behaviour of the old CLI that does not know about this query parameter.


dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
```diff
@@ -262,7 +262,7 @@ public void testDepsCommandWithUnknownRunners() throws IOException {
     @Test
     public void testDepsCommand() throws IOException {
         Client.main(new String[] { "--config", TestUtility.getConfigFileLocation(true), "deps" });
-        Assert.assertTrue(systemOutRule.getLog().contains("monotonic=="));
+        Assert.assertFalse(systemOutRule.getLog().contains("monotonic=="));
```
denis-yuen, line 265: It seems like this would pass if the log just had an empty string. 
If monotonic is no longer a dependency, this should test a different dependency

denis-yuen (commented):
>  I set it to null instead of false/true to mimic the behaviour of the old CLI that does not know about this query parameter. 

To confirm understanding, is null being serialized to false or just omitted, meaning that these are becoming soft refreshes?

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
garyluu: Null is ommited. If it was false, it would outright fail some tests. These are all becoming hard refreshes because the default in the web service is true.

monotonic is the difference between python 3 and python 2.  Python 2 has it, python 3 doesn't. I've changed the default returned by this command to python 3 so it would not have it.

The test would not pass for a empty string because of the assertion function on the next line.

denis-yuen: > The test would not pass for a empty string because of the assertion function on the next line. 

Ah, thanks for the explanation. 

If asserting that monotonic is not installed is a way of ensuring that the environment is using Python 3 rather than Python2, add that intention as a code comment.

denis-yuen (changes requested):
> Null is ommited. If it was false, it would outright fail some tests. These are all becoming hard refreshes because the default in the web service is true.

Ah, so null is becoming true in these cases. 

Can we change this to true in general and then refactor one or two tests to run again using "null" instead of true to mimic the old CLI labelled as a regression test?  That would be a bit more intuitive for me. 

Good idea to test the behavior of the old CLI though

denis-yuen:
Much closer, would still like the 

> refactor one or two tests to run again using "null" instead of true

It was a good idea, just not immediately intuitive

garyluu:
> Much closer, would still like the
> 
> > refactor one or two tests to run again using "null" instead of true
> 
> It was a good idea, just not immediately intuitive

Still unclear.  Is dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowRegressionIT.java not what you wanted?

denis-yuen (approved):
> Still unclear. Is dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowRegressionIT.java not what you wanted?

Duh, github was only showing changes since last review. Cool

coverbeck (approved):
It's unfortunate that the Swagger generator doesn't also generate a method where you don't need to specify the default parameters. I guess that could get messy though.

NatalieEO (approved):

PR #57, opened 2021-01-28
test 1.10.1

NatalieEO:

- [] Check that you pass the basic style checks and unit tests by running `mvn clean install`


THIRD-PARTY-LICENSES.txt
```diff
@@ -120,12 +120,12 @@ Lists of 362 third-party dependencies.
...
-     (Apache Software License, Version 2.0) dockstore-client (io.dockstore:dockstore-client:1.10.0-beta.0-SNAPSHOT - https://github.com/dockstore/dockstore-cli/dockstore-client)
-     (Apache Software License, Version 2.0) dockstore-common (io.dockstore:dockstore-common:1.10.0 - https://github.com/dockstore/dockstore/dockstore-common)
-     (Apache Software License, Version 2.0) dockstore-file-plugin-parent (io.dockstore:dockstore-file-plugin-parent:1.10.0-beta.0-SNAPSHOT - https://github.com/dockstore/dockstore-cli/dockstore-file-plugin-parent)
-     (Apache Software License, Version 2.0) dockstore-language-plugin-parent (io.dockstore:dockstore-language-plugin-parent:1.10.0 - https://github.com/dockstore/dockstore/dockstore-language-plugin-parent)
-     (Apache Software License, Version 2.0) dockstore-webservice (io.dockstore:dockstore-webservice:1.10.0 - https://github.com/dockstore/dockstore/dockstore-webservice)
+     (Apache Software License, Version 2.0) dockstore-cli-integration-testing (io.dockstore:dockstore-cli-integration-testing:1.11.0-alpha.0-SNAPSHOT - https://github.com/dockstore/dockstore-cli/dockstore-cli-integration-testing)
```
coverbeck, line 123: This seems wrong. Did you branch off develop and are merging to master?

NatalieEO: No, I made a new branch off of master. Just now locally I changed it to 1.10.1 on master and did a build. The licenses changed to 1.11.0 as well

coverbeck: OK, I see, it looks like master has 1.11.0-alpha.SNAPSHOT as the version. It should be 1.10.1-SNAPSHOT, I think.

I think we need to do the whole mvn release process for CLI?

NatalieEO: Not sure, The checklist just says to test it. @denis-yuen ?

denis-yuen: The CLI is always a bit of a judgement call. If the CLI depends on a new webservice or has a new feature, then its worth a release. Keep in mind that we've gone a number of releases without a synchronised CLI release. 
Otherwise, we're just testing whether the new webservice is horribly breaking something. 

In this case, I don't see anything that should affect the CLI https://github.com/dockstore/dockstore/compare/1.10.0...1.10.1 and the CLI doesn't seem to have anything new https://github.com/dockstore/dockstore-cli/pulls?q=is%3Apr+is%3Aclosed so I'd just see the passing tests below and call it a day on this checklist item.

THIRD-PARTY-LICENSES.txt
```diff
@@ -120,12 +120,12 @@ Lists of 362 third-party dependencies.
...
+     (Apache Software License, Version 2.0) dockstore-client (io.dockstore:dockstore-client:1.11.0-alpha.0-SNAPSHOT - https://github.com/dockstore/dockstore-cli/dockstore-client)
+     (Apache Software License, Version 2.0) dockstore-common (io.dockstore:dockstore-common:1.10.1 - https://github.com/dockstore/dockstore/dockstore-common)
+     (Apache Software License, Version 2.0) dockstore-file-plugin-parent (io.dockstore:dockstore-file-plugin-parent:1.11.0-alpha.0-SNAPSHOT - https://github.com/dockstore/dockstore-cli/dockstore-file-plugin-parent)
+     (Apache Software License, Version 2.0) dockstore-language-plugin-parent (io.dockstore:dockstore-language-plugin-parent:1.10.1 - https://github.com/dockstore/dockstore/dockstore-language-plugin-parent)
+     (Apache Software License, Version 2.0) dockstore-webservice (io.dockstore:dockstore-webservice:1.10.1 - https://github.com/dockstore/dockstore/dockstore-webservice)
```
denis-yuen, line 128: Note the webservice that this is using seems to be 1.10.1

THIRD-PARTY-LICENSES.txt
coverbeck: I keep asking this every release, and I think the answer is no, but do we need to create a 1.10.1 CLI release so that the CLI will download correctly.

denis-yuen: IIRC (which is probably not): newer `dockstore` scripts download from `dockstore/dockstore-cli`. Older scripts download from `dockstore/dockstore`. jars look in those locations as well. 

ping @wshands

wshands: Both the dockstore scripts on dockstore/dockstore 1.10.0 and dockstore/dockstore-cli 1.10.0 have this version of the CLI specified:
DEFAULT_DOCKSTORE_VERSION="1.10.0"
So that is the CLI release that  will be downloaded. If a dockstore/dockstore (webservice) 1.10.1 is created, then the dockstore/dockstore dockstore script must have the 1.10.0 version of the CLI specified because that is the current CLI version. When we fix the link to the dockstore script on https://dockstore.org/quick-start and we are convinced most users are using a version of CLI > 1.8.0 we can remove the dockstore script from dockstore/dockstore.
If an old CLI (<= 1.8.0 ) is being used and the CLI update command is executed it will erroneously report that the latest CLI version is 1.10.1, because it checks the dockstore/dockstore release versions assuming they are the same as CLI versions. This was fixed in CLI 1.9.0 so the CLI points to the CLI repo to check for the latest or unstable versions.

coverbeck: So, to confirm, this PR is really to make sure the tests pass, but we're not actually going to do a release off of it.

It feels a little odd, because master will now contain non-released code, but I guess master never lines up exactly anyway.

denis-yuen (approved):
With the caveat that this doesn't get reverted and is just testing

garyluu (approved):

PR #58, opened 2021-02-04
Feature/tranistion to circleci

wshands:
Enables unit and integration tests on CircleCI


.circleci/config.yml
```diff
@@ -1,44 +1,297 @@
...
+      - run:
+          name: build
+          command: mvn -B clean install -DskipTests
+
+        # Running scan must occur after build
+      - run_git_secrets_scan
+      - check_license_file
+      - run:
+          name: copy m2 manually
+          command: cp -r ~/.m2 .
```
coverbeck, line 209: There is already code that saves and restores the .m2 directory, e.g., line 196. Can you use that instead?

.circleci/config.yml
```diff
@@ -47,17 +300,105 @@ jobs:
...
+            echo "export SINGULARITY_PATH=${HOME}/go" >> $BASH_ENV
+            echo "export PATH=${GO_PATH}/go/bin:${PATH}:${SINGULARITY_PATH}/bin" >> $BASH_ENV
+      - run:
+          name: install pip dependencies
+          command: scripts/install-tests.sh
+      - install_confidential_test_data
+
+  run_tests:
+    steps:
+      - run:
```
coverbeck, line 385: Was this a debugging step you left in?

.circleci/config.yml
```diff
@@ -1,44 +1,297 @@
...
+      TESTING_PROFILE: unit-tests
+    steps:
+      - setup_for_tests
+      - run_tests
+      - clean_and_save_cache
+      - save_test_results
+      - send_coverage
+
+  singularity-tests:
+    # Must use machine instead of Docker for Singularity
```
denis-yuen, line 90: Nicely documented

.circleci/config.yml
```diff
@@ -1,44 +1,297 @@
...
+      - setup_for_tests
+      - setup_machine_env_vars
+      - setup_for_integration_tests
+      - run_tests
+      - clean_and_save_cache
+      - save_test_results
+      - send_coverage
+
+
+    # This catagory seems to be covered by confidential and non-confidential tests
```
denis-yuen, line 162: Can delete and seems reasonable that it would be covered by confidential vs non-confidential

.circleci/config.yml
```diff
@@ -47,17 +300,105 @@ jobs:
...
+          name: setup postgres user and DB
+          command: |
+            psql -c "create user dockstore with password 'dockstore' createdb;" -U postgres
+            psql -c 'create database webservice_test with owner = dockstore;' -U postgres
+            psql -c "ALTER USER dockstore WITH superuser;" -U postgres
+
+  install_postgres_sql_client:
+    steps:
+     - run:
+         name: Install postgresql client
```
denis-yuen, line 336: Think this means the postgresql client is always the latest, might have been like this before but is something to be aware of

denis-yuen (approved):
Seems to pass a smell check looking at numbers of tests in circle ci. 
Singularity is taking 15 minutes to run 2 tests which might be worth a follow-up ticket.

.circleci/config.yml
```diff
@@ -1,44 +1,297 @@
...
+      - setup_machine
+      - setup_for_tests
+      - setup_machine_env_vars
+      - setup_for_integration_tests
+      - run_tests
+      - clean_and_save_cache
+      - save_test_results
+      - send_coverage
+
+  non-confidential-tests:
```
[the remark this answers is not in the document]
denis-yuen: Good idea, this syntax might help https://github.com/dockstore/dockstore/blob/develop/.circleci/config.yml#L31

.circleci/config.yml
wshands: Added a comment in the code

.circleci/config.yml
wshands: Deleted commented out code

.circleci/config.yml
wshands: Removed

.circleci/config.yml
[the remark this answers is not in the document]
wshands: Consolidated job code.

.circleci/config.yml
wshands: Not sure I will have to investigate.

.circleci/config.yml
```diff
@@ -1,44 +1,214 @@
...
+          POSTGRES_HOST_AUTH_METHOD: trust
+          POSTGRES_PASSWORD: postgres
+      - image: docker.elastic.co/elasticsearch/elasticsearch:7.10.0
+        environment:
+          - xpack.security.enabled: false
+          - transport.host: localhost
+          - network.host: 127.0.0.1
+          - http.port: 9200
+          - discovery.type: single-node
+    resource_class: medium+
```
coverbeck, line 25: If you don't need Postgres/ElasticSearch, see previous comment, then this probably doesn't need to be a medium+

.circleci/config.yml
coverbeck: I feel this is worth resolving -- we should only do it one way if possible.

coverbeck (changes requested):

.circleci/config.yml
```diff
@@ -1,44 +1,214 @@
...
+  build-tools: circleci/build-tools@2.7.0
+executors:
+  tests_executor:
     docker: # run the steps with Docker
-      - image: circleci/openjdk:11.0.4-jdk-stretch
+      - image: cimg/openjdk:11.0.8
         environment:
           PGHOST: 127.0.0.1
-      - image: circleci/postgres:9.6.5-alpine-ram
+      - image: circleci/postgres:11.8
```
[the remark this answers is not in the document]
wshands: Removed ES for unit tests

.circleci/config.yml
```diff
@@ -1,44 +1,214 @@
-version: 2
-jobs:
-  build: # runs not using Workflows must have a `build` job as entry point
-    resource_class: medium+
+version: 2.1
+orbs:
+  build-tools: circleci/build-tools@2.7.0
+executors:
+  tests_executor:
```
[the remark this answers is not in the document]
wshands: I renamed the executor.

.circleci/config.yml
wshands: Removed the resource_class

.circleci/config.yml
wshands: Now .m2 is only saved to cache (when calling save_cache)

.circleci/config.yml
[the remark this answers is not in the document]
coverbeck: My comment was for Postgres and ES; you only removed ES, pretty sure you can remove PG as well.

coverbeck (changes requested):
This looks really good, but I still think you can get rid of Postgres when running the unit tests.

.circleci/config.yml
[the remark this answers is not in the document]
wshands: Removed Postgres from unit tests

GFJHogue (approved):
Any reason why empty file `circle_ci_test_data.zip.enc` is added?

coverbeck (approved):

scripts/decrypt.sh
```diff
@@ -1,15 +1,23 @@
...
+: "$CIRCLE_CI_KEY"
+: "$CIRCLE_CI_IV"
+
+openssl aes-256-cbc -d -in circle_ci_test_data.zip.enc -k "$CIRCLE_CI_KEY" -iv "$CIRCLE_CI_IV" -out secrets.tar
+tar xvf secrets.tar
+# Create the directory where the .pem file will be placed for continuous integration
+# If you would like to change this path (might be a good idea since we no longer use
+# TravisCI for CI) you need to change the path in dockstoreTest.yml which currently
+# is in the encrypted bundle circle_ci_test_data.zip.enc
+sudo mkdir -p /home/travis
```
garyluu, line 20: Shouldn't this be changed to w/e natalie used?

wshands: Yes but the path is located in the encrypted bundle; there is no partialDockstoreTest.yml as in dockstore/dockstore; but maybe I'll just go ahead an fix it in the encrypted bundle.

wshands: I fixed it

PR #69, opened 2021-05-03
remove regression test elements

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-2475
Regression tests are run from dockstore/dockstore to test older CLI versions against newer webservice master or release branches. Regression tests do not need to be in the CLI project as we are not testing newer CLI versions against older webservice versions.


coverbeck (approved):

NatalieEO (approved):

garyluu (approved):

Richard-Hansen (approved):

.circleci/config.yml
```diff
@@ -221,6 +221,9 @@ commands:
          name: Install postgresql client
          command: |
            sudo rm -rf /var/lib/apt/lists/*
+           # heroku.com is not updating their keys
```
denis-yuen, line 224: Any idea why this is not needed in dockstore/dockstore or is this about to fail there too?
`cimg/openjdk` is slightly older but the machine image appears to be the same

denis-yuen (approved):

.circleci/config.yml
wshands: I do not know why this is not needed in dockstore/dockstore

PR #71, opened 2021-06-25
CLI Release 1.11.0 rc.0 changes to develop

wshands:


denis-yuen (commented):
Changes looked reasonable but there looked like relevant build issues to resolve

coverbeck (commented):
Was here, seconding Denis' comment on build issues.

coverbeck (commented):
Shouldn't there be a release/1.11.0 branch, with a PR against that?

coverbeck:
> Shouldn't there be a release/1.11.0 branch, with a PR against that?

On further reflection, I guess if nothing has changed on develop, it will work.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -303,11 +303,11 @@ private Workflow getDockstoreWorkflowByPath(String path) {
         // simply getting published descriptors does not require permissions
         Workflow workflow = null;
         try {
-            workflow = workflowsApi.getPublishedWorkflowByPath(path, null, false, null);
+            workflow = workflowsApi.getPublishedWorkflowByPath(path, "versions", false, null);
```
coverbeck, line 306: This means that the 1.10 client will break against the 1.11 web service, doesn't it? I don't think there's much to do about it at this point.

coverbeck (approved):

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
wshands: Yes

NatalieEO (approved):

Richard-Hansen (approved):

denis-yuen (approved):

PR #75, opened 2021-07-09
Update bom which also updates jackson cbor

denis-yuen:
https://ucsc-cgl.atlassian.net/browse/SEAB-2642

Note: I think the change to the `getWorkflowByPath` endpoint also means that the 1.12 webservice breaks compatibility with 1.11 clients. Please verify @NatalieEO 
Wonder if it would be possible to have it check for boolean as a fallback to 1.11 behaviour


coverbeck (approved):
Agreed that it would be good not to break 1.11 client with 1.12 release.

garyluu (approved):

denis-yuen:
> Agreed that it would be good not to break 1.11 client with 1.12 release.

Split out https://ucsc-cgl.atlassian.net/browse/SEAB-3126 due to OOO

NatalieEO (approved):

PR #81, opened 2021-08-30
Feature/3159/nightly deploy

garyluu:
For dockstore/dockstore#3159 See counterpart: https://github.com/dockstore/dockstore/pull/4430

Added maven wrapper using: `mvn -N io.takari:maven:wrapper -Dmaven=3.6.3`

Before merging, need to:
- ~change it so that it runs on cron on develop branch only~
- ~Need to change snapshot location to w/e (whether it's GitLab or OICR Artifactory, etc)~
- Maybe remove the git diff line (although might be useful)

Possible enhancements:
GitHub action is currently hardcoded to use `-Ddockstore-core.version=1.12.0-alpha.1-SNAPSHOT`. This will not work once dockstore/dockstore changes to some other snapshot version of a different name.  Might be a way to make this better

See the same but different snapshot version working in https://github.com/dockstore/dockstore-cli/actions
Irrelevnt commit to show JCommander back to 1.78: Has JCommander with 1.78
Don't use cached snapshots, show diff: Has JCommander with 1.81


.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright 2007-present the original author or authors.
```
denis-yuen, line 2: I don't think this needs to be checked in. The webservice repo should have the minimum needed

mvnw.cmd
```diff
@@ -0,0 +1,182 @@
+@REM ----------------------------------------------------------------------------
```
denis-yuen, line 1: this is for windows (doesn't need to be checked in)

denis-yuen (commented):
> GitHub action is currently hardcoded to use -Ddockstore-core.version=1.12.0-alpha.1-SNAPSHOT. This will not work once dockstore/dockstore changes to some other snapshot version of a different name. Might be a way to make this better

One snapshot per day might be ok anyway
i.e. since we know when the cron ran, we know which snapshot jar was used

.github/workflows/mvnw.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v2
+      # Step that does that actual cache save and restore
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+           ${{ runner.os }}-maven-
+      - name: Build with mvnw
+        run: ./mvnw clean install -DskipTests -Ddockstore-core.version=1.12.0-alpha.1-SNAPSHOT -B -ntp -U
```
coverbeck, line 23: 1. Don't we want to run the tests, or is this just a compiler check?
2. Not 100% sold on this idea, but what if the dockstore-core.version value came from an environment variable?

garyluu: 1. Probably going to need to ask Walt about this, but it was my assumption that compiler check would've caught most if not all the issues that appeared during release 
2. Hmm...not sure. I think for GitHub admins it would be easier with environment variable since you don't have to do a PR (which avoids wait time and such). But for non-admins, it's either much more difficult or impossible

.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1,2 @@
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.3/apache-maven-3.6.3-bin.zip
```
denis-yuen, line 1: Could probably use a newer version like 3.8.1
https://github.com/dockstore/dockstore/blob/develop/.mvn/wrapper/maven-wrapper.properties

Is 3.6.3 on purpose?

.github/workflows/mvnw.yml
denis-yuen: Might as well run unit tests to avoid using the really crummy builds

denis-yuen (commented):
> GitHub action is currently hardcoded to use `-Ddockstore-core.version=1.12.0-alpha.1-SNAPSHOT`. This will not work once dockstore/dockstore changes to some other snapshot version of a different name.  Might be a way to make this better

Create a ticket for this, will probably break or get neglected soon
https://jfrog.com/knowledge-base/how-to-find-the-latest-artifact-version-based-on-layout/
https://stackoverflow.com/questions/47742074/how-to-find-the-latest-artifact-version-with-artifactory-query-language

.mvn/wrapper/maven-wrapper.properties
garyluu: yes, i originally tried to use the newest version.  There were some differences in the generated pom.xml.  This version is the version from the image.

denis-yuen: Hmmm, please create a ticket for this as well and assign to me. 
I'm using Maven 3.8.1 locally and don't see any differences in the generated pom (Java 11.0.11)

denis-yuen (commented):
(re: retro discussion item) 
please link follow-up tickets

NatalieEO (approved):

coverbeck (approved):

.github/workflows/mvnw.yml
```diff
@@ -0,0 +1,25 @@
+name: Java CI
+
+#on:
```
denis-yuen, line 3: Intentionally commented? 
Will this run on every build without this?

pom.xml
```diff
@@ -55,10 +56,20 @@
     </scm>
 
     <repositories>
+        <repository>
+            <id>gitlab-maven</id>
```
denis-yuen, line 60: we're not really using gitlab

denis-yuen (approved):
some minor things to look at

.github/workflows/mvnw.yml
garyluu: one of the last things to switch out before merging. it's still commented out for PR testing purposes

.github/workflows/mvnw.yml
```diff
@@ -0,0 +1,23 @@
+name: Java CI
+
+on:
+  schedule:
+    - cron: '0 16 * * *'
```
denis-yuen, line 5: https://crontab.guru/#0_16_*_*_* has this as every 4pm, think this might be a noon-ish deploy and not a nightly deploy. But :shrug:

denis-yuen (approved):

PR #83, opened 2021-09-13
Update mvnw and try snapshot query

denis-yuen:
Working on https://github.com/dockstore/dockstore/issues/4448 and  https://github.com/dockstore/dockstore/issues/4449
Main repo is already up-to-date actually, updating wrapper 
Using query to get latest timestamped version and then trim away the timestamp (which seems weird, there is probably a better way to do this)

Standardizing with Maven 3.8.1 and Java 11.0.11

To see how this works with the nightly build see https://github.com/dockstore/dockstore-cli/pull/83/commits/acdbff97127124fb2eccc48337a7991cf2dd45b9 ("Looks like openjdk image has old maven" has the working github action cron build) Following commits should only affect circle ci and not github actions (which is used for the cron test of the cli)

For future updates, the Go failure on the update looks like it is due to conflicting Go installations with the new Circle CI image 
https://github.com/ledgerwatch/erigon/issues/1518

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 executors:
   unit_tests_executor:
     docker: # run the steps with Docker
-      - image: cimg/openjdk:11.0.10
+      - image: cimg/openjdk:11.0.11
 
   machine_integration_test_exec:
     machine: # run the steps with Ubuntu VM
@@ -81,7 +81,7 @@ jobs:
 
   build:
     docker: # run the steps with Docker
-      - image: cimg/openjdk:11.0.10
+      - image: cimg/openjdk:11.0.11
         environment:
           # Java can read cgroup. Sadly the cgroup in
           # CircleCI is wrong. Have to manually set. Nothing to do with surefire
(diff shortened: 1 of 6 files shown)


coverbeck (approved):

NatalieEO (approved):

garyluu (commented):
Double-checking that the following is true:
- Nothing is using the latest maven (3.8.2)
- Maven wrapper is going to use 3.8.1
- CircleCI is using maven 3.6.3
- Enforcer is enforcing 3.6.3

If so, what's the rationale for creating this discrepancy? The issue was for dockstore/dockstore, this PR seems to introduce the issue to dockstore/dockstore-cli

denis-yuen:
> If so, what's the rationale for creating this discrepancy? The issue was for dockstore/dockstore, this PR seems to introduce the issue to dockstore/dockstore-cli

A fair question. I think I misread the issue as being dockstore being inconsistent with dockstore-cli as opposed to inconsistent Maven versions. Re-working.

garyluu (approved):

PR #84, opened 2021-10-01
Service-info endpoint support for WES

Richard-Hansen:
Adds support for the `/service-info` WES endpoint (https://ga4gh.github.io/workflow-execution-service-schemas/docs/#operation/GetServiceInfo). This will make testing initial AGC work easier. 

Command: `dockstore workflow wes service-info`

The Dockstore-cli help command looks like:
Screen Shot 2021-10-01 at 12 06 53 PM

And the response printed looks like: 
Screen Shot 2021-10-01 at 12 06 22 PM
or
Screen Shot 2021-10-01 at 1 51 50 PM

(Note: the null responses are dependent on what the WES endpoint provides, and the generated ServiceInfo class explicitly prints `class ServiceInfo` as part of the info dump, which seems odd.)

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -67,6 +67,7 @@ import io.openapi.wes.client.api.WorkflowExecutionServiceApi;
 import io.openapi.wes.client.model.RunId;
 import io.openapi.wes.client.model.RunLog;
 import io.openapi.wes.client.model.RunStatus;
+import io.openapi.wes.client.model.ServiceInfo;
 import io.swagger.client.ApiException;
 import io.swagger.client.model.Label;
 import io.swagger.client.model.SourceFile;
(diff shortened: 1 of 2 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1429,6 +1443,17 @@ private void wesCancelHelp() {
         printHelpFooter();
     }
 
+    private void wesServiceInfoHelp() {
+        printHelpHeader();
+        out("Usage: dockstore " + getEntryType().toLowerCase() + " wes service-info --help");
```
coverbeck, line 1448: The entry type is irrelevant for this endpoint.

But it opens a can of worms because `wes` always follows the entry type, so leave it for now.

coverbeck (approved):

garyluu (approved):

denis-yuen (commented):
Seems like you could test that the help messages come up with no exceptions and clear out some of these codecov messages

denis-yuen (approved):

PR #92, opened 2021-10-27
Feature/3437/trs url body

Richard-Hansen:
Adds support for passing the primary descriptor URL under the `workflow_url` as a TRS URL, rather than passing it inlined as a `workflow_attachment`.

This was split out from https://github.com/dockstore/dockstore-cli/pull/91, which addresses AWS SigV4 signing of HTTP payloads. The current AWS WES service we have setup does not accept the workflow as a `workflow_attachment`, which served as motivation to adding this PR's feature for TRS URL passing under `workflow_url`.

---

**Note**: I'm hoping for feedback specific to the calculation of the TRS URL, as I'm not a fan of the current implementation. Is there a better way to build the TRS URL?


wshands (approved):

garyluu (approved):

Richard-Hansen:
> As a follow-up enhancement, can you please create a ticket to not download the entry at all for the TRS URL case (assuming I'm right about that).
> 
> Might take a major code restructuring, but we shouldn't be downloading the entry if we're not going to do anything with it.

Created a ticket: https://github.com/dockstore/dockstore/issues/4557

PR #98, opened 2021-11-22
Feature/3663/WesWithJCommander

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-3663

This PR is the first half of another ticket, https://ucsc-cgl.atlassian.net/browse/DOCK-1985. In this PR, we add the framework for the Dockstore CLI to use jCommander for WES calls, rather than manual parsing. The actual use of this framework comes into play in the partner PR: https://github.com/dockstore/dockstore-cli/pull/99.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesCommandParser.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesCommandParser.java
@@ -42,20 +42,20 @@ public class WesCommandParser {
     @Parameters(separators = "=", commandDescription = "Execute WES commands")
     public static class WesMain {
         @Parameter(names = "--wes-url", description = "The URL of the WES server.", required = false)
-        private static String wesUrl = null;
+        private String wesUrl = null;
         @Parameter(names = "--wes-auth",
             description = "The authorization type and value of this wes request. --wes-auth <type> <value>. "
                 + "Type can be 'bearer' or 'aws'. "
                 + "Value can be a token if the type is 'bearer', or an AWS profile if the type is 'aws'",
             variableArity = true,
             required = false)
-        private static List<String> wesAuth = null;
+        private List<String> wesAuth = null;
         @Parameter(names = "--aws-config", description = "A path to an AWS configuration file containing AWS profile credentials.", required = false)
-        private static String awsConfig = null;
+        private String awsConfig = null;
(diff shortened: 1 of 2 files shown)


dockstore-client/src/test/java/io/dockstore/client/cli/WesCommandParserTest.java
```diff
@@ -0,0 +1,372 @@
...
+        assertNull("The parsed auth type should be 'null'", wesCommandParser.wesMain.getWesAuthType());
+        assertNull("The parsed auth value should be null", wesCommandParser.wesMain.getWesAuthValue());
+    }
+
+    @Test
+    public void testCommandLaunchHelp() {
+        final String[] args = {
+            "--wes-auth",
+            "bearer",
+            "123456",
```
denis-yuen, line 122: Making a note in the other PR, but approving for now

denis-yuen (approved):

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesCommandParser.java
```diff
@@ -0,0 +1,155 @@
...
+        @Parameter(names = "--wes-auth",
+            description = "The authorization type and value of this wes request. --wes-auth <type> <value>. "
+                + "Type can be 'bearer' or 'aws'. "
+                + "Value can be a token if the type is 'bearer', or an AWS profile if the type is 'aws'",
+            variableArity = true,
+            required = false)
+        private List<String> wesAuth = null;
+        @Parameter(names = "--aws-config", description = "A path to an AWS configuration file containing AWS profile credentials.", required = false)
+        private String awsConfig = null;
+        @Parameter(names = "--aws-region", description = "The AWS region of the WES server.", required = false)
```
coverbeck, line 53: 1. I'm not convinced the Dockstore CLI should have this argument. I'd say force the user to put the region in the AWS profile.
2. If we keep it, `... of the WES server` part of the message should not be generic like that; it's only going to apply to AGC, and the message should reflect that.

I haven't looked at your other PR yet, so we can carry this on there, and I'll approve.

coverbeck (approved):

wshands (approved):

PR #99, opened 2021-11-23
Feature/1985/modify wes credential passing

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/DOCK-1985
Currently targeted at branch in this PR: https://github.com/dockstore/dockstore-cli/pull/98, but is probably easiest to re-target to develop after #98  is merged.

Allows the Dockstore CLI to accept AWS profiles for credentials, and modifies the way some credentials are handled. If you wanted to run a command entirely from the command line, it would look something like:
```
dockstore workflow wes launch \
--wes-url <url> \
--wes-auth <type> <authorization> \
--entry myEntry \
--json myJson
```
You can also specify your configuration within the Dockstore CLI config file:
```
[WES]
url: https://1234567.execute-api.us-east-1.amazonaws.com/prod/ga4gh/wes/v1
type: aws
authorization: wes-test-profile
```
Which reduces what you need to pass in on the command line to:
```
dockstore workflow wes launch \
--entry myEntry \
--json myJson
```

Additional Notes:
1. Originally, the CLI config WES section took `authorization: Bearer <token>` as a parameter, which is now broken into 2 lines: 
```
type: Bearer
authorization: <token>
```
3. Added a note to https://ucsc-cgl.atlassian.net/browse/SEAB-3665 to have the BOM specify CDK and AWS SDK versions.
4. There's a Snyk finding as well, looks to be from the AWS SDK version. There is a conflict in upgrading the AWS SDK version with a version we get from the webservice, so perhaps it's worthwhile to finish the BOM ticket before this one?


pom.xml
```diff
@@ -145,6 +145,12 @@
                 <artifactId>jaxb-api</artifactId>
                 <version>2.3.1</version>
             </dependency>
+            <!-- https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk -->
+            <dependency>
+                <groupId>com.amazonaws</groupId>
+                <artifactId>aws-java-sdk-simpleworkflow</artifactId>
+                <version>1.11.86</version>
```
garyluu, line 152: if the aws-java-sdk-core and this is likely to remain the same version, use a property

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1525,11 +1572,9 @@ private void wesServiceInfoHelp() {
     private void printWesHelpFooter() {
         out("Global Optional Parameters:");
         out("  --wes-url <WES URL>                 URL where the WES request should be sent, e.g. 'http://localhost:8080/ga4gh/wes/v1'");
-        out("  --wes-auth <auth>                   Authorization credentials for the WES endpoint, e.g. 'Bearer 12345'");
-        out("  --aws                               Flag indicating whether the WES request will be to an AWS endpoint");
-        out("  --aws-access-key <key>              The AWS access key associated with an IAM user/profile.");
-        out("  --aws-secret-key <key>              The AWS secret key associated with an IAM user/profile.");
-        out("  --aws-region <region>               The AWS region the WES endpoint is located in, e.g. 'us-east-1'.");
+        out("  --wes-auth <authType> <authValue>   Authorization credentials for the WES endpoint, e.g. 'Bearer 12345' or 'aws myProfile'");
```
denis-yuen, line 1575: The vote is a little unclear from the storytime, but would recommend not allowing bearer via command-line
(Also note discussion around https://app.slack.com/archives/CKREJ1938/p1637341883053900)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1226,22 +1229,66 @@ public WesRequestData aggregateWesRequestData(final List<String> args) {
...
+        final String wesEndpointUrl = ObjectUtils.firstNonNull(
+            command.getWesUrl(),
+            Objects.requireNonNull(configSubNode).getString("url"));
+
+        // Determine the authorization method used by the user
+        final String authType = ObjectUtils.firstNonNull(
+            command.getWesAuthType(),
+            Objects.requireNonNull(configSubNode).getString("type"));
+
+        // The auth value is either a bearer token or AWS profile
```
denis-yuen, line 1247: See comment later

denis-yuen (commented):
Not sure if bearer tokens is being addressed elsewhere

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1177,45 +1138,87 @@ private void wesServiceInfo(WorkflowExecutionServiceApi clientWorkflowExecutionS
         }
     }
 
+    /**
+     * Given the parsed command object, determine if we are to print help commands
+     * @param wesCommandParser Parse commands
+     * @return
```
coverbeck, line 1144: If you keep the signature this way, fill out the `@return`

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1226,22 +1229,66 @@ public WesRequestData aggregateWesRequestData(final List<String> args) {
...
+
+            String accessKey = null;
+            String secretKey = null;
+
+            // If the user specified AWS as their intended target, but provided no profile name, assume they don't want to pass credentials
+            // in this request.
+            if (authValue != null) {
+
+                try {
+                    // Get the AWS config path
```
coverbeck, line 1264: I commented elsewhere, but I'm not convinced we need this. One could say it doesn't hurt since you already added it, but I do worry about all the command-line options.

Also, if I'm wrong and it's a fairly common practice, then perhaps it's fine.

pom.xml
Richard-Hansen: Added a property: https://github.com/dockstore/dockstore-cli/pull/99/commits/c1864bef14187d41e149c5e582cff59b82c007de

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: Brainstorming:
We could allow tokens via the command line, but encourage using a config file via our documentation. This would be good for the users who want the bear-bones approach, even though it isn't best practice. 

Or, If we don't want to allow `--wes-auth bearer <token>` at all, it seems like we could drop the  entire`--wes-auth` parameter, as the only remaining valid option would be `--wes-auth aws <profile>`. Dropping `--wes-auth` would mean all credentials are required to be passed in via the config file (until AWS environment variables are fully supported.)

Both approaches seem reasonable, thoughts?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1177,45 +1138,87 @@ private void wesServiceInfo(WorkflowExecutionServiceApi clientWorkflowExecutionS
         }
     }
 
+    /**
+     * Given the parsed command object, determine if we are to print help commands
+     * @param wesCommandParser Parse commands
+     * @return
+     */
+    private boolean displayWesHelp(WesCommandParser wesCommandParser) {
```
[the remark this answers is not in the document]
Richard-Hansen: Good points, renamed: https://github.com/dockstore/dockstore-cli/pull/99/commits/3bb5f2c817276fb79599208aea88474729fba840

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1226,22 +1229,66 @@ public WesRequestData aggregateWesRequestData(final List<String> args) {
...
-        final String wesEndpointUrl = ObjectUtils.firstNonNull(wesUri, Objects.requireNonNull(configSubNode).getString("url"));
+        // Obtain the WES command object
+        JCommander parsedCommand = wesCommandParser.jCommander
+            .findCommandByAlias(wesCommandParser.jCommander.getParsedCommand());
+        WesCommandParser.WesMain command = parsedCommand == null ?  new WesCommandParser.WesMain() : (WesCommandParser.WesMain) parsedCommand.getObjects().get(0);
+
+        // Attempt to find the WES URL
+        final String wesEndpointUrl = ObjectUtils.firstNonNull(
+            command.getWesUrl(),
+            Objects.requireNonNull(configSubNode).getString("url"));
```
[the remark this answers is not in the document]
Richard-Hansen: Fixed: https://github.com/dockstore/dockstore-cli/pull/99/commits/b3bbaaff5d6e9504fd9bfd8d33c178ae01397b0a

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: Updated: https://github.com/dockstore/dockstore-cli/pull/99/commits/3bb5f2c817276fb79599208aea88474729fba840

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: Removed. See comment above.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1525,11 +1572,9 @@ private void wesServiceInfoHelp() {
...
         out("Global Optional Parameters:");
         out("  --wes-url <WES URL>                 URL where the WES request should be sent, e.g. 'http://localhost:8080/ga4gh/wes/v1'");
-        out("  --wes-auth <auth>                   Authorization credentials for the WES endpoint, e.g. 'Bearer 12345'");
-        out("  --aws                               Flag indicating whether the WES request will be to an AWS endpoint");
-        out("  --aws-access-key <key>              The AWS access key associated with an IAM user/profile.");
-        out("  --aws-secret-key <key>              The AWS secret key associated with an IAM user/profile.");
-        out("  --aws-region <region>               The AWS region the WES endpoint is located in, e.g. 'us-east-1'.");
+        out("  --wes-auth <authType> <authValue>   Authorization credentials for the WES endpoint, e.g. 'Bearer 12345' or 'aws myProfile'");
+        out("  --aws-region <region>               An AWS region, e.g. 'us-east-1', only required for requests to AWS WES servers.");
+        out("  --aws-config <config>               Path to an AWS credentials file. Defaults to ~/.aws/credentials");
```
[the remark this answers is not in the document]
Richard-Hansen: I ended up removing the `--aws-config` and `--aws-region` command line options, and am reading both of those from the AWS config files: `~/.aws/config` and `~/.aws/credentials`, which is the default behavior provided from the AWS SDK. This means that all AWS credentials will be fully specified in AWS resources, and the only value given to Dockstore is the profile name.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1226,22 +1229,66 @@ public WesRequestData aggregateWesRequestData(final List<String> args) {
...
+                    // 1) The path to the config file is invalid or 2) the profile name is invalid or 3) The config file is malformed
+                    errorMessage(e.getMessage(), CLIENT_ERROR);
+                }
+
+            }
+
+            // Get the AWS region we are send the request to
+            final String region = ObjectUtils.firstNonNull(
+                command.getAwsRegion(),
+                Objects.requireNonNull(configSubNode).getString("region"));
```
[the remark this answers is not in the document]
Richard-Hansen: Removed `--aws-region` from the command line and `region` from the config file. See other comment https://github.com/dockstore/dockstore-cli/pull/99#discussion_r758823959

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1226,22 +1229,66 @@ public WesRequestData aggregateWesRequestData(final List<String> args) {
...
-            // TODO should probably have the config sections all defined within a single class. At least have enums.
-            // AWS credentials shouldn't be stored in the dockstore config, so we wont check for it there.
-            final String accessKey = optVal(args, "--aws-access-key", null);
-            final String secretKey = optVal(args, "--aws-secret-key", null);
-            final String region = optVal(args, "--aws-region", null);
+
+            String accessKey = null;
+            String secretKey = null;
+
+            // If the user specified AWS as their intended target, but provided no profile name, assume they don't want to pass credentials
```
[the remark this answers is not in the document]
Richard-Hansen: Added. Now, if no profile is passed, we search for `default`.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1226,22 +1229,66 @@ public WesRequestData aggregateWesRequestData(final List<String> args) {
...
+                } catch (IllegalArgumentException | SdkClientException e) {
+                    // Some potential reasons for this exception are:
+                    // 1) The path to the config file is invalid or 2) the profile name is invalid or 3) The config file is malformed
+                    errorMessage(e.getMessage(), CLIENT_ERROR);
+                }
+
+            }
+
+            // Get the AWS region we are send the request to
+            final String region = ObjectUtils.firstNonNull(
```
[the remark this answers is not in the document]
Richard-Hansen: The region is now being pulled from the AWS profile by the AWS SDK. `--aws-region` is removed from the command line as well. More details here: https://github.com/dockstore/dockstore-cli/pull/99#discussion_r758823959.

Richard-Hansen:
Just for some extra details. AWS maintains 2 separate credential files: `~/.aws/config` and `~/.aws/credentials`. The `config` file is designated for non-confidential values (region, display settings, etc...) and `credentials` is for keys and tokens (This isn't a strict requirement, but seems to be the expected practice, the 2 files can overlap in information).

At the time of this comment, the current CLI implementation doesn't allow specifying a custom path to either file, and they are expected to be at  `~/.aws/config` and `~/.aws/credentials`. The AWS Java SDK handles the pathing for us, so all we need to provide is a profile name. By letting the AWS SDK handling the file location, and not allowing custom pathing, this allows us to remove `--aws-config`, `--aws-credentials`, and `--aws-region` command line and config file parameters. We could re-add this functionality down-the-line, if it seems worthwhile, but for the moment reducing the extra complexity appears reasonable.

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
wshands: If a user wanted to run workflows on different WES server providers (maybe unrealistic) and script the process it would be a lot more difficult without the --wes-auth parameter?

wshands (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesRequestData.java
```diff
@@ -70,8 +70,12 @@ public WesRequestData(String url, String awsAccessKey, String awsSecretKey, Stri
 
         // The credentials that were passed in are null, so we are assuming they are within an authorized environment (such as an EC2 instance)
```
coverbeck, line 71: This comment no longer seems applicable here. Can you please move it to where it is applicable, if I'm reading this right.

Richard-Hansen: Reworded: https://github.com/dockstore/dockstore-cli/pull/99/commits/e4fe3cc94e2f5bf1587faa9f23acc57afd92b334

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesRequestData.java
```diff
@@ -68,8 +68,7 @@ public WesRequestData(String url, String awsAccessKey, String awsSecretKey, Stri
         this.awsSecretKey = awsSecretKey;
         this.region = region;
 
-        // The credentials that were passed in are null, so we are assuming they are within an authorized environment (such as an EC2 instance)
-        // This also assumes there is no scenario where just one of the keys is sufficient.
+        // AWS credentials are required to make a complete API call. If any of the parameters are null, throw an error for the user.
```
coverbeck, line 71: Just to clarify, the no credentials case (instance profile, behind a firewall, etc.) will still work, right? I guess that's a different path; you just wouldn't specify auth at all?

Richard-Hansen: This blocks the no credential case, as an error will be thrown noting the user is missing either keys or a region if any inputs are null. I think the initial implementation of 'no credentials' was flawed.

If I'm understanding the documentation correctly, we should be able to setup a proper credential parsing chain using the AWS SDK. That chain will parse the EC2 instance metadata for IAM credentials (See [InstanceProfileCredentialsProvider](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/InstanceProfileCredentialsProvider.html)). 

AWS provides a large number of default credential providers, each parsing credentials from different locations and in different orders: [AWSCredentialsProvider](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/AWSCredentialsProvider.html), and we can chain together the ones we want.

That's a long way around of saying that for this ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-3478, we should be able to add another credential provider that explicitly searches for EC2 instance credentials, and passes those to the `WesRequestData` object as normal. Because of this, if `WesRequestData` receives null AWS credentials as inputs, throwing an error seems reasonable.

coverbeck: > This blocks the no credential case, as an error will be thrown noting the user is missing either keys or a region if any inputs are null.

The no-credentials case will still work for non-AWS WES, right? E.g., I'm running Cromwell locally as a WES Server and don't set up any authentication for it...

I think it's fine to hold off on the instance provider case, since you have the followup ticket.

Richard-Hansen: Ah, yes, no-credentials for non-AWS WES still works. `WesRequestData` accepts null bearer tokens, just not null AWS credentials.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
denis-yuen: Still feels weird although I recognize the pain of having to keep editing. 
Maybe should just drop it

denis-yuen (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: Decided to remove it for the time being: https://github.com/dockstore/dockstore-cli/pull/99/commits/dc431f8ae47cc5f5f4f25bfd369f58453540e630, we can re-add `--wes-auth` down the line if we want.

coverbeck (approved):

PR #100, opened 2021-12-08
Feature/3709/workflow attachments and params

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-3709

This PR addresses some changes to the Dockstore CLI's interpretation of the WES schema. Specifically, more support for the `workflow-attachment` property has been added, such that parameter files can be passed according to the WES server's specific implementation. We are also treating `workflow_params` as a JSON object (not a file) when sending it in the HTTP request. Although odd, this seems to match up with the WES spec, Toil, and AGC. Just not the WES client library we are using. The above ticket has some more details as to why this is required. 

This PR removes the `--json`, `--yaml` and `--uuid` parameters for the WES launch commands, as they shouldn't be required. The `--input` and `--attach`/`-a` parameters were added and work as follows:
1. `--input` is a path to a JSON file, and sets the `workflow-params` property of the WES request
2. `--attach` is a path to any file, and appends it to the `workflow-attachment` property. You can attach as many or as few files as you want.

The old `WESLauncher.java` class was a little convoluted and passed through various other 'launch' infrastructure that wasn't used. So, a new replacement class `WesLauncher.java` is added in this PR which is directly called when the user wants to execute a WES launch command. 

What isn't supported for this PR:
1. Launching local files to a WES endpoint (We should be able to hook into existing provisioning code to add this)
2. Automatically pulling attachment files from the Dockstore entry 
3. Allowing users to attach a directory with `--attach` (perhaps a new switch for that? e.g. `--attach-directory`)
4. Path expansion, (e.g. `~/my/file.json`)

Because we are no longer going through other launch functions, we aren't provisioning files locally, so this ticket may also be able to close: https://ucsc-cgl.atlassian.net/browse/DOCK-1977


coverbeck:
Initial feedback based on reading your PR description only, I haven't looked at your code:

> This PR removes the --json, --yaml and --uuid parameters for the WES launch commands, as they shouldn't be required. The --input and --attach/-a parameters were added and work as follows:

1. I'm too lazy to look it up; what is the `--uuid` parameter?
2. Off-hand, I don't like removing `--json` and `--yaml`. Running a workflow with the CLI should as much as possible the same experience when running a workflow locally vs. on WES. If I run a workflow locally with `dockstore workflow launch ... --json foo.json`, I would expect to pass foo.json the same way when running against a WES endpoint. Still wrapping my head around it, but what if we kept `--json` and `--yaml`, and dropped `--input`.
3. Regarding `--attach`, what sort of files would one attach beyond test parameters?

> 1. Launching local files to a WES endpoint (We should be able to hook into existing provisioning code to add this)

What do you mean by local files? The descriptors or actual data files? The provisioning code is only for data, so I'm a little confused.
> 2.  Automatically pulling attachment files from the Dockstore entry
Do you mean the test parameter files or something else? This would be nice, but we should do it for local workflows as well if/when we do it.

Richard-Hansen:
> 1. I'm too lazy to look it up; what is the `--uuid` parameter?

Its description was "Allows you to specify a uuid for 3rd party notifications". This parameter was never actually used for WES, the issue was that the WES launch command was passed through the standard jCommander launch, and so it had to be added to avoid an error. This is one of the reasons the WES launch being split from the standard launch is nice.

> 2. Off-hand, I don't like removing `--json` and `--yaml`. Running a workflow with the CLI should as much as possible the same experience when running a workflow locally vs. on WES. If I run a workflow locally with `dockstore workflow launch ... --json foo.json`, I would expect to pass foo.json the same way when running against a WES endpoint. Still wrapping my head around it, but what if we kept `--json` and `--yaml`, and dropped `--input`.

Based on the WES schema, I don't think `--yaml` is a valid option (unless we want to transpile input yaml to json?). We could change `--input` -> `--json` though.

> 3. Regarding `--attach`, what sort of files would one attach beyond test parameters?

I'm thinking this mainly comes into play for local launches. Currently, we are only passing a TRS URL to the primary descriptor, and we hope the WES server will resolve all secondary descriptors relative to that URL. But, for when we support local entries, the secondary descriptors will need to be manually added as attachment files.

This plays into the TODO enhancement of uploading an entire directory with the `--attach-directory` parameter (name not set in stone), as it will allow all primary, secondary and test parameters files in a directory to be easily attached to the WES request.

> What do you mean by local files? The descriptors or actual data files? The provisioning code is only for data, so I'm a little confused.

Both. For example, you could attach your primary descriptor file in the request with `--attach`, and set the `workflow_url` (`--entry` parameter) to its relative path, rather than a TRS URL (this is currently supported, but should be). So the CLI command would be something like:
```
dockstore workflow wes launch --entry Dockstore.wdl --attach /path/to/primary/Dockstore.wdl
```
And that ^ is a valid way to execute a single-descriptor workflow. Another way to think about the `workflow_url`, `workflow_params` and `workflow_attachment` WES request properties is:
1. `workflow_attachment` contains all primary, secondary, and test parameter files.
2. `workflow_url` references a file in `workflow_attachment`
3. `workflow_params` references a file in `workflow_attachment`

> Do you mean the test parameter files or something else? This would be nice, but we should do it for local workflows as well if/when we do it.

As a possible UX enhancement, it seems like it would be really nice for a user to run:
```
dockstore workflow wes launch --entry my/entry/workflow:v1
```
And we attach the test parameter json, and (potentially) other descriptors required to launch the workflow. That way, they don't need to re-specify the test parameter file that may already displayed in Dockstore.


---
To clarify the WES functionality scope (based on the past storytime), we have `3` methods we want to support for launching workflows with WES.
1. The entry is on Dockstore. The primary descriptor is referenced by a TRS URL, and the WES server resolves all dependencies using that URL. Parameter files may be passed in manually.
2. The user manually attaches local files, which includes all files needed to execute the workflow (No need for anything registered on Dockstore). This means that all files are included as attachments in the WES request, and the WES server only looks at files provided in the HTTP request to execute the workflow.
3. The user specifies an entry on Dockstore, and we provision the necessary files locally so the WES request will look like method 2. (This is for the case where the user has their entry on Dockstore, as in case 1, but the WES server does not support TRS URLs to resolve dependencies, and so the request needs to look like case 2.)

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,200 @@
...
+        File workflowParams = loadFile(workflowParamPath);
+
+        // A list of supplementary files that are required to run the workflow. This may include any/all of the following:
+        // 1. The primary descriptor file
+        // 2. Secondary descriptor files
+        // 3. Test parameter files
+        // 4. Any other files referenced by the workflow
+        // TODO: Allow users to specify a directory to upload?
+        // TODO: Automatically attach all files referenced in remote Dockstore entry?
+        List<File> workflowAttachment = loadAttachments(attachments);
```
wshands, line 61: Are the changes to workflow_params and workflow_attachment going to work on the other WES server implementations, e.g. Arvados, DNAStack, DNANexus, Illumina, etc.? Some of these WES implementations may no longer exist or may not be implemented correctly, but also we may not want to break compatibility with those that do exist? I recommend especially confirming that these changes are compatible with DNAStack and Patrick Magee <patrick@dnastack.com> and how he understands the schema. In addition there is a GA4GH ticket to remove the workflow_attachment from the WES schema [here](https://github.com/ga4gh/workflow-execution-service-schemas/issues/124) along with a discussion about alternatives. I advise trying to keep options open (not sure how exactly) since the schema seems to be in flux for quite some time now.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,200 @@
...
+        return new File(workflowParamPath);
+    }
+
+    /**
+     * Given a list of string paths, this attempts to convert it to a list of files.
+     *
+     * @param attachments A list of paths that correspond to files that need to be attached to the WES request
+     * @return A list of File objects
+     */
+    public static List<File> loadAttachments(List<String> attachments) {
```
wshands, line 166: There can be files other than descriptors that are imported by the workflow, e.g. $import and $include in CWL. I don't know whether TRS will retrieve those files along with secondary descriptors.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
wshands: Keep an eye on the issues and discussions [here](https://github.com/ga4gh/workflow-execution-service-schemas/issues)

coverbeck:
FYI, my general approach here is that we should hide WES details from the end user as much as possible. But maybe that will prove wrong.
> 
> Based on the WES schema, I don't think `--yaml` is a valid option (unless we want to transpile input yaml to json?). We could change `--input` -> `--json` though.

Ah, I see. I'd assumed because Cromwell and CWL support YAML, but this is tricker via WES. I'd vote to change `--input` to `--json` and not do any conversions.

> > 3. Regarding `--attach`, what sort of files would one attach beyond test parameters?
> 
> I'm thinking this mainly comes into play for local launches. Currently, we are only passing a TRS URL to the primary descriptor, and we hope the WES server will resolve all secondary descriptors relative to that URL. But, for when we support local entries, the secondary descriptors will need to be manually added as attachment files.

The CLI can parse a workflow client-side and determine which descriptors are used. But maybe is should pass the entire directory. Maybe we could do this via the existing `--local-entry`? E.g., if that is specified, just upload the entire directory?

> > What do you mean by local files? The descriptors or actual data files? The provisioning code is only for data, so I'm a little confused.
> 
> Both. For example, you could attach your primary descriptor file in the request with `--attach`, and set the `workflow_url` (`--entry` parameter) to its relative path, rather than a TRS URL (this is currently supported, but should be). So the CLI command would be something like:
> 
> ```
> dockstore workflow wes launch --entry Dockstore.wdl --attach /path/to/primary/Dockstore.wdl
> ```

We already have `dockstore workflow launch --local-entry /path/to/primary/Dockstore.wdl`, it would be nice to have the one way for both.

> > Do you mean the test parameter files or something else? This would be nice, but we should do it for local workflows as well if/when we do it.
> 
> As a possible UX enhancement, it seems like it would be really nice for a user to run:
> 
> ```
> dockstore workflow wes launch --entry my/entry/workflow:v1
> ```

It's a little more complex than this because there can be more than one test parameter file. Or none. But something like this would be nice; right now you have to manually download the parameter file.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,200 @@
...
+        if (attachments == null) {
+            return null;
+        }
+
+        List<File> workflowAttachments = new ArrayList<>();
```
coverbeck, line 171: This is fine, but if interested, you can do this with `map` as well.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,200 @@
...
+
+        // TODO: Handle path expansions? i.e. '~/my/file.txt'
+        return new File(workflowParamPath);
+    }
+
+    /**
+     * Given a list of string paths, this attempts to convert it to a list of files.
+     *
+     * @param attachments A list of paths that correspond to files that need to be attached to the WES request
+     * @return A list of File objects
```
coverbeck, line 164: or null (but see my next comment)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,200 @@
...
+     */
+    public static List<File> loadAttachments(List<String> attachments) {
+        if (attachments == null) {
+            return null;
```
coverbeck, line 168: As a general rule of thumb, I would return an empty list instead of null. Then the caller doesn't have to do null checks.

I haven't read all your code yet, so maybe there's a good reason for it in this case, but FYI.

Also for your consideration, my other rule of thumb is if it can return null, then return Optional. I know it's verbose, but it forces the caller to check. We don't do this in a lot of places, so just if you want.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
coverbeck, line 166: Naming nitpicking:

1. I would rename `attachments` to `filePaths` or `paths`, because that's what they are.
2. Not wild about `loadAttachments` -- you're just creating `File` objects here.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,200 @@
...
+            escapedId,
+            escapedVersionId,
+            escapedType,
+            escapedRelativePath);
+    }
+
+    /**
+     * Attempts to load a file from the provided path. Errors out if the file doesn't exist.
+     *
+     * @param workflowParamPath Path to a file
```
coverbeck, line 142: or null

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,200 @@
...
+     * @return A File object
+     */
+    public static File loadFile(String workflowParamPath) {
```
coverbeck, line 145: I find `load` in the method name confusing. It implies to me that you're reading its contents.

Richard-Hansen:
> Ah, I see. I'd assumed because Cromwell and CWL support YAML, but this is tricker via WES. I'd vote to change `--input` to `--json` and not do any conversions.

Changed: https://github.com/dockstore/dockstore-cli/pull/100/commits/a05599fe9fd1f144d398abddf2a823a71d04cc4c

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Fixed: https://github.com/dockstore/dockstore-cli/pull/100/commits/277163d31739c42fafc725eeca20c959a72ddffc

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Returning an empty list: https://github.com/dockstore/dockstore-cli/pull/100/commits/277163d31739c42fafc725eeca20c959a72ddffc

Also added Optionals

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Renamed to `filePaths`, `fetchFiles()` and `fetchFile()` (https://github.com/dockstore/dockstore-cli/pull/100/commits/277163d31739c42fafc725eeca20c959a72ddffc). 

I'm actually not certain what the proper convention is for "A file object I'm creating now that will get read by something later". If `fetch` is still a little iffy, let me know :)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Fixed: https://github.com/dockstore/dockstore-cli/pull/100/commits/277163d31739c42fafc725eeca20c959a72ddffc

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Changed, see comment above: https://github.com/dockstore/dockstore-cli/pull/100/commits/277163d31739c42fafc725eeca20c959a72ddffc

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: This actually wont work with DNAStack. We made a change a while back for TRS URLs that replaced the local file attachment approach. This ticket is to add back that functionality: https://ucsc-cgl.atlassian.net/browse/DOCK-1981.
We could potentially retarget these WES changes to a large feature branch until everything is ready to be moved to develop?

Thanks for pointing out the possible WES changes, I'll keep an eye on that.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: In that case, they'd need to be attached here as well I think. I'm actually not certain how a WES server would decide to switch between resolving by TRS URL and/or using the list of attached files, but seems to be an implementation-specific scenario.

Either way, we should have support for multiple file-upload methods once https://ucsc-cgl.atlassian.net/browse/DOCK-1981 is complete.

Richard-Hansen:
> The CLI can parse a workflow client-side and determine which descriptors are used. But maybe is should pass the entire directory. Maybe we could do this via the existing `--local-entry`? E.g., if that is specified, just upload the entire directory?

When `--local-entry` is re-added (https://ucsc-cgl.atlassian.net/browse/DOCK-1981), I think we may want to have uploading the entire directory as a flag? I could see a user having a bunch of files in a directory and just wanting to upload one as a local entry. I'll link to your original comment chain in the above ticket.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1079,8 +1079,9 @@ WorkflowExecutionServiceApi getWorkflowExecutionServiceApi() {
      * This will attempt to launch a workflow given the command arguments
      *
      */
-    private void wesLaunch(final String entry, final String localEntry, final String jsonRun, final String yamlRun, final String wdlOutput, final boolean ignoreChecksumFlag, final String uuid) {
-        launchWithArgs(entry, localEntry, jsonRun, yamlRun, wdlOutput, ignoreChecksumFlag, uuid);
+    private void wesLaunch(String workflowEntry, String workflowParams, List<String> workflowAttachment) {
+        // Only supports workflows at the moment
+        WesLauncher.launchWesCommand((WorkflowClient) this, workflowEntry, workflowParams, workflowAttachment);
```
coverbeck, line 1084: You don't have to change it, but I don't like that downcast to WorkflowClient; it could be missed when/if we support tools. I.e., you're relying on codeflow to never hit this when it's a ToolClient; you could instead have the compiler do the check for you.

A better way might be to:

1. Have this  method always throw an error and override it WorkflowClient
1. Make this method abstract, then implement it in the subclasses, where the ToolClient implementation is to error out.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesCommandParser.java
```diff
@@ -55,12 +57,10 @@ public boolean isHelp() {
...
         @Parameter(names = "--entry", description = "Complete workflow path in Dockstore (ex. NCI-GDC/gdc-dnaseq-cwl/GDC_DNASeq:master)", required = true)
         private String entry;
-        @Parameter(names = "--json", description = "Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs")
+        @Parameter(names = "--json", description = "JSON file describing which attached file contains input parameters.")
         private String json;
-        @Parameter(names = "--yaml", description = "Parameters to the entry in Dockstore, one map for one run, an array of maps for multiple runs")
-        private String yaml;
-        @Parameter(names = "--uuid", description = "Allows you to specify a uuid for 3rd party notifications")
-        private String uuid;
+        @Parameter(names = {"--attach", "-a"}, description = "A list of paths to files that should be included in the WES request", variableArity = true)
```
coverbeck, line 62: How are multiple files separated? By comma, space, or something else? I actually don't know :), so it would be good to add it to the description.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+            LOG.error("Error launching WES run", e);
+        }
+    }
+
+    /**
+     * Obtains the workflow on Dockstore
+     *
+     * @param workflowClient The WorkflowClient
+     * @param workflowEntry The workflow entry (i.e. github.com/myRepo/myWorkflow:version)
+     * @return A Workflow object for the corresponding workflow+version
```
coverbeck, line 95: What happens if the entry doesn't exist? Throws an exception or returns null? If null, please update this; if an exception, I'd add it to line 91 (even though we don't always document runtime exceptions).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+     * Given a list of string paths, this attempts to convert it to a list of files.
+     *
+     * @param filePaths A list of paths that correspond to files that need to be attached to the WES request, this may be null
+     * @return A list of File objects, which may be empty
+     */
+    public static List<File> fetchFiles(List<String> filePaths) {
+
+        // Return an empty list if no attachments were passed
+        if (filePaths == null) {
+            return new ArrayList<>();
```
coverbeck, line 171: Tip, fine the way you have it, but you can return `Collections.emptyList()` -- it's more for the use case where a method is called thousands or millions of times, so that you're not creating objects that need to be garbage collected, but since this is called once, it doesn't matter, but just for future reference.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+     * @return An Optional File object, this may be empty if the filepath is null
+     */
+    public static Optional<File> fetchFile(String filePath) {
+        if (filePath == null) {
+            return Optional.empty();
+        }
+
+        // Verify the file path exists
+        Path path = Paths.get(filePath);
+        if (!Files.exists(path)) {
```
coverbeck, line 153: I think, but you should verify, that  `Files.exists` return true if it's a directory (I know that's true for `File.exists`). If yes, do you want this?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: Followed approach 2. Added an implementation to `ToolClient` and `WorkflowClient`, the `ToolClient` implementation throws an error: https://github.com/dockstore/dockstore-cli/pull/100/commits/8bf0df02649d0da6dc572bb27c3bafd0a8c2a861

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesCommandParser.java
Richard-Hansen: Files are separated by spaces, or the parameter can be used multiple times. The following are equivalent:
1. `-a <file1> <file2> <file3>`
2. `-a <file1> -a <file2> -a <file3>`

Updated the parameter description and help command: https://github.com/dockstore/dockstore-cli/pull/100/commits/3e058383cb1567468bdcbb089b54981fd24095e9

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: The `WorkflowClient` throws an exception and the program exits if the entry doesn't exist.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Interesting, I hadn't considered that case, thanks for the pointer.

Updated: https://github.com/dockstore/dockstore-cli/pull/100/commits/09bf7c4684c788157d1b3e3f8838a7cca44b246c

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Good catch, `File.exists` returned true for both files and directories, which caused a NPE down the line. 

Updated to `File.isRegularFile` which explicitly checks if it is a file, also added a test to check the directory case: https://github.com/dockstore/dockstore-cli/pull/100/commits/b6b66faa279194b7955db8111f745ef4cc623df5

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
coverbeck: It looks like an uncaught exception, which is why the program exits. This can be fine, but we generally catch these (see line 86).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * Obtains the workflow on Dockstore
+     *
+     * @param workflowClient The WorkflowClient
+     * @param workflowEntry The workflow entry (i.e. github.com/myRepo/myWorkflow:version)
+     * @return A Workflow object for the corresponding workflow+version
+     */
+    public static Workflow getWorkflowForEntry(WorkflowClient workflowClient, String workflowEntry) {
```
coverbeck, line 98: I suggest adding `throws APIException` to the method signature.

coverbeck (approved):
I don't feel too strongly about either of my suggestions.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Added: https://github.com/dockstore/dockstore-cli/pull/100/commits/8167b87d138806083bad92f5cffede2a8bafb2f2

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Oops, I miswrote. An we raise an error for the user, rather than throw an exception. This occurs here: https://github.com/dockstore/dockstore-cli/blob/8167b87d138806083bad92f5cffede2a8bafb2f2/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L320

So, in this case, there is no exception just an intentional `System.exit()` and message.

denis-yuen (approved):

PR #108, opened 2022-01-18
minor refactor

Richard-Hansen:
Minor refactoring for file downloads. The CLI was returning the expected path to downloaded primary descriptors (in temporary directories) regardless of whether or not the file actually exists. In the case where we do not automatically unzip the directory, the primary descriptor would not be present, so we should return the existing zip file instead.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
@@ -377,10 +377,13 @@ public class WorkflowClient extends AbstractEntryClient<Workflow> {
             workflowsApi.getApiClient().setDebugging(Client.DEBUG.get());
             File zipFile = new File(directory, zipFilename(workflow));
             FileUtils.writeByteArrayToFile(zipFile, arbitraryURL, false);
+
+            // If we unzip the file, we can provide a path to the primary descriptor, otherwise just provide a path to the zip file
             if (unzip) {
                 SwaggerUtility.unzipFile(zipFile, directory);
+                return new File(directory, first.get().getWorkflowPath());
             }
-            return new File(directory, first.get().getWorkflowPath());
+            return zipFile;
         } else {
             throw new RuntimeException("version not found");
         }


dockstore-cli-integration-testing/generated/src/main/resources/pom.xml
```diff
@@ -46,8 +46,8 @@
       <scope>compile</scope>
       <exclusions>
         <exclusion>
```
denis-yuen, line 48: think the proper order is actually group then artifact, may need to update Maven and/or use mvnw

denis-yuen (approved):

dockstore-cli-integration-testing/generated/src/main/resources/pom.xml
Richard-Hansen: This is the ordering produced from `mvnw`, so perhaps we need to bump the wrapper's version?

garyluu: weird, i used mvnw on develop and it does have these changes

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -377,10 +377,13 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
...
             File zipFile = new File(directory, zipFilename(workflow));
             FileUtils.writeByteArrayToFile(zipFile, arbitraryURL, false);
+
+            // If we unzip the file, we can provide a path to the primary descriptor, otherwise just provide a path to the zip file
             if (unzip) {
                 SwaggerUtility.unzipFile(zipFile, directory);
+                return new File(directory, first.get().getWorkflowPath());
             }
-            return new File(directory, first.get().getWorkflowPath());
+            return zipFile;
```
coverbeck, line 386: This comment for line 354 -- JavaDoc should reflect that it can return a path to the zip file.

dockstore-cli-integration-testing/generated/src/main/resources/pom.xml
denis-yuen: Worth a shot. A little puzzled why it was the other order in the first place then, but maybe this repo doesn't have the "fail on changes" check that the main repo does

coverbeck: This is odd; we ran into it, I forget which way, when Walt made the dockstore-cli for the hotfix. I don't know what's going on -- what plugin actually does the sorting?

coverbeck (commented):
> so we should return the existing zip file instead.

What is the existing zip file? Is it guaranteed to exist?

Richard-Hansen:
> What is the existing zip file? 

The zip file will contain the workflow's associated files, such as descriptors (one or more) and test parameter files. (Here is the endpoint https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowZip)

For example, a basic workflow would produce a zip file containing just `Dockstore.wdl`, on the other hand a multi-descriptor workflow would produce a zip file containing something like `Dockstore.wdl`, `Secondary.wdl` and `testParams.json`. 

> Is it guaranteed to exist?

The code checks to ensure the workflow is valid [here](https://github.com/dockstore/dockstore-cli/blob/a522fbda5ee24a99a29d4a85a34d40a56609eb54/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L367), which indicates at least one descriptor is present and obtainable to add to the zip file.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
Richard-Hansen: Added: https://github.com/dockstore/dockstore-cli/pull/108/commits/2ff233d41b72c36e144948f8cbe7271760a5965a

coverbeck:
> > What is the existing zip file?
> 
> The zip file will contain the workflow's associated files, such as descriptors (one or more) and test parameter files. (Here is the endpoint https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowZip)
> 

Sorry, I wasn't clear at all. I know what the ZIP file is, but when I asked that question, I was thinking about the WES case where we don't pass the descriptors themselves, but their TRS URLs. In that case, we don't need to download the zip file at all (I think you have a ticket to optimize that?). Will this code get invoked in that case? If yes, what happens when the zip file hasn't been downloaded?

Maybe I'm jumping ahead to the optimization ticket.

Richard-Hansen:
> Sorry, I wasn't clear at all. I know what the ZIP file is, but when I asked that question, I was thinking about the WES case where we don't pass the descriptors themselves, but their TRS URLs. In that case, we don't need to download the zip file at all (I think you have a ticket to optimize that?). Will this code get invoked in that case? If yes, what happens when the zip file hasn't been downloaded?
> 
> Maybe I'm jumping ahead to the optimization ticket.

Ah, makes sense, good question. I am currently working on the ticket that allows for uploading your Dockstore entry as if it were locally stored on your machine, and this current PR supports that goal.

**To answer the core question**: No, the CLI will not download any files unless the user explicitly adds a flag to their command.

**A little more in-depth**: The current design is to add a flag `--provision-locally`, which will indicate to the CLI if it should/shouldn't download the descriptor files. If `--provision-locally` is set, the CLI will call the download function that is being modified in this PR, and receive a path back to the downloaded zip file. This zip file will be subsequently unzipped into another nested temp directory, and all contents of the nested directory will be attached to the WES request (This is why this PR is necessary, as I need the path to the zip file, not the primary descriptor).

coverbeck (approved):

dockstore-cli-integration-testing/generated/src/main/resources/pom.xml
denis-yuen: Should be Maven itself, looks like 3.8.2 or above

PR #107, opened 2022-01-14
Feature/seab 3772/update webservice

garyluu:
Updating the webservice lead to a lot of test failures:

- New Nextflow version observed to require the NXF_HOME environment variable to be set
- SHA1 -> SHA256 checksum change and the use of generated column requires quite a few changes
- Public Amazon ECR tool was prevented in the webservice

Updated the postgres version to match develop webservice too

I removed a chunk of the checksum tests...not quite sure how to go about doing it in a better way

For SEAB-3772

PR #111, opened 2022-01-20
API parameter is required to be non-null.

Richard-Hansen:
Followup to: https://github.com/dockstore/dockstore-cli/pull/107

This parameter is now required to be non-null. There are two separate 'bioworkflow' strings to choose from:
1. `WorkflowSubClass.BIOWORKFLOW.toString()` -> `"BIOWORKFLOW"`
2. `WorkflowClient.BIOWORKFLOW` -> `"bioworkflow"`

I see `WorkflowSubClass.BIOWORKFLOW.toString()` being used to make the same function call [elsewhere](https://github.com/dockstore/dockstore-cli/pull/107/files#diff-e679fea995a50bafa2ef51b5d382128204370e4a29ef6509d0e2271a89dc3981), so I'm opting for it here, but both are used throughout the CLI. Either seems to work. Is there an important difference between the two? I believe both are processed as case-insensitive.

This parameter

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
@@ -12,6 +12,7 @@ import java.util.Optional;
 
 import io.dockstore.openapi.client.ApiClient;
 import io.dockstore.openapi.client.ApiException;
+import io.dockstore.openapi.client.model.WorkflowSubClass;
 import io.openapi.wes.client.model.RunId;
 import io.swagger.client.model.Workflow;
 import io.swagger.client.model.WorkflowVersion;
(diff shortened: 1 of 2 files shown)


coverbeck (approved):
> I see WorkflowSubClass.BIOWORKFLOW.toString() being used to make the same function call elsewhere, so I'm opting for it here, but both are used throughout the CLI. Either seems to work. Is there an important difference between the two? I believe both are processed as case-insensitive.

I would have thought it's case-sensitive, but to my surprise [DropWizard doc](https://www.dropwizard.io/en/latest/manual/validation.html#enum-constraints) says it's case-insensitive, and I verified that. I prefer the way you have it now.

Added Natalie as a reviewer since she worked on the input.

NatalieEO (approved):

denis-yuen (approved):

PR #113, opened 2022-01-21
Feature/1981/local file provisioning

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/DOCK-1981

We have 3 separate methods for launching a workflow using WES with the Dockstore CLI:
1. Pass a TRS URL to the primary descriptor to the WES server, and let the WES server resolve all dependencies **(Complete)**
2. Download a Dockstore entry locally and upload all the files in the HTTP request **(This PR)**
3. Launch from files that are not on Dockstore, and just kept locally **(TODO in a future ticket)**

---
This PR re-introduces local provision for the Dockstore CLI's launch functionality. This means that the CLI will download a Dockstore entry to a temporary directory, and then attach the downloaded files to the WES request. This is achieved by adding the `--provision-locally` flag to your request. If `--provision-locally` is not set, the CLI will pass the TRS URL of the primary descriptor to the WES server, and not download any files.

This was tested against a DNAStack WES server, which raised some issues with the way WES handles file attachments, mainly:

The WES API does not maintain directory structures when sending files as part of an HTTP request. This means that when a file is attached to the WES HTTP request, the Dockstore CLI must decide what `filename` to attach to the file. This causes issues when descriptors reference files in a relative fashion (say via imports), as there can be multiple different relative paths to the same file, but we can only give each file a single name. (This would all be solved if we could send the entire workflow as a zipped bundle, that way the WES server would be able to unpack it and maintain the directory structure, but that doesn't appear to match the WES spec).

To address the file naming issue, a new class `WesFile` was created, which extends `File`, and calculates a custom filename based on `prefix`/`suffix` values. This causes the following:
1. If files are downloaded to a temporary directory, such as to `/tmp/directory/nested/file.wdl`, where `/tmp/directory/` is the tmp directory and `nested/file.wdl` is the path in the GitHub repository, the HTTP request will send this file named as `nested/file.wdl`.
2. If a user attaches a file using a relative path, such as `--attach relative/path/file.txt`, the HTTP request will send this file named as `relative/path/file.txt`
3. If the user attaches a file using an absolute path, such as `--attach /absolute/path/file.txt`, this will throw an error, as we don't know the proper file naming.
4. Descriptor imports will fail if they are relative to a descriptor, rather than the repo root. For example, given the directory structure:
```
tmp/
 \_ dir/
      \_ Dockstore.wdl
      \_ nested/
           \_ secondary.wdl
```
The HTTP request will upload files named `dir/Dockstore.wdl` and `dir/nested/secondary.wdl`. If `Dockstore.wdl` expects there to be a file named `nested/secondary.wdl`, it will fail. We could name files relative to the primary descriptor, but this has two main drawbacks:
1. Relative imports may still fail for secondary descriptors, as files are named relative to the primary descriptor only.
2. WES does not allow directory traversal, i.e. `../file.json`, meaning that all attached files must be in an equal or lower directory than the primary descriptor.

Any opinions or suggestions on the above approach are appreciated, there are multiple ways we could implement local file provisioning.


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1515,7 +1516,9 @@ private void printWesLaunchHelpBody() {
         out("");
         out("Optional parameters:");
         out("  --json <json file>                  JSON parameter file for the WES run. This may be reference an attached file");
-        out("  --attach <path, -a <path>           A list of paths to files that should be included in the WES request. (ex. -a <path1> <path2> OR -a <path1> -a <path2>)");
+        out("  --attach <path>, -a <path>          A list of paths to files that should be included in the WES request. (ex. -a <path1> <path2> OR -a <path1> -a <path2>)");
```
denis-yuen, line 1519: Is this what was referred to by `(This would all be solved if we could send the entire workflow as a zipped bundle, that way the WES server would be able to unpack it and maintain the directory structure, but that doesn't appear to match the WES spec).` 

Referencing each file this way seems like a real pain. Maybe a new ticket could be to attach files by zip bundle but have the CLI unpack it and essentially use `--attach` under the hood on each file in it. Not sure if that makes sense, so wait for user feedback?

denis-yuen: Or maybe an option to point at a directory recursively instead?

denis-yuen (commented):
Not sure if comment makes sense, would be new ticket anyway

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: There is a TODO (https://github.com/dockstore/dockstore-cli/pull/113/files#diff-c271a67a154db53da69eacd582cb800f031f4b3f259d437fa22bdaebc0255d82L76) to allow users to specify a directory to attach and the CLI will find and attach all files to the WES request.

What I meant by: 

> (This would all be solved if we could send the entire workflow as a zipped bundle, that way the WES server would be able to unpack it and maintain the directory structure, but that doesn't appear to match the WES spec).

Is that if WES supported receiving a single zip file which contained all attachment files, and then on the WES server itself the zip files were unpacked, we wouldn't have this directory structure issue. (AGC supports sending zip files, but you must also supply a `MANIFEST.json` that points to it and the primary descriptor). But, since this isn't the current state of WES, it doesn't affect this PR.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesCommandParser.java
```diff
@@ -60,6 +60,8 @@ public boolean isHelp() {
     public static class CommandLaunch extends WesMain {
         @Parameter(names = "--entry", description = "Complete workflow path in Dockstore (ex. NCI-GDC/gdc-dnaseq-cwl/GDC_DNASeq:master)", required = true)
         private String entry;
+        @Parameter(names = "--provision-locally", description = "Indicates that the provided entry should be downloaded locally and pass inline to the WES server.")
```
coverbeck, line 63: For the CLI, I associate "provision" with provisioning files (our file provisioning plugins). so I feel the word is getting a bit overloaded. 

Also, the key point here is the request to WES; one request has a reference to the descriptor, the other request has the descriptors inline. How we create an inline request (by downloading files) is arguably an implementation detail.

So, how about instead of `--provision-locally`, something like `--inline-descriptors` or `--inline-workflow`, and the description only talks about how the request is formed?

This is a proposal I'm not sure of, and willing to be talked out of.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -117,6 +134,54 @@ public static Workflow getWorkflowForEntry(WorkflowClient workflowClient, String
...
+            exceptionMessage(ex, "A problem was encountered while downloading the entry.", IO_ERROR);
+            throw new RuntimeException(ex);
+        }
+
+        // The directory where the unzipped workflow contents are located
+        File unzippedWorkflowDir;
+
+        // Unzip the workflow to a nested temporary directory
+        try {
+            unzippedWorkflowDir = Files.createTempDirectory(Path.of(tempDir.getAbsolutePath()), DOCKSTORE_NESTED_TEMP_DIR_PREFIX).toFile().getAbsoluteFile();
```
coverbeck, line 174: Add a line `unzippedWorkflowDir.deleteOnExit()` to clean up.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -117,6 +134,54 @@ public static Workflow getWorkflowForEntry(WorkflowClient workflowClient, String
...
+     * @param descriptorType The type of descriptor for this workflow (WDL, CWL, etc...)
+     * @return A zipped File object
+     */
+    public static File provisionFilesLocally(WorkflowClient workflowClient, String workflowEntry, String descriptorType) {
+
+        // A temporary directory which will house all downloaded content
+        File tempDir;
+
+        try {
+            tempDir = Files.createTempDirectory(DOCKSTORE_ROOT_TEMP_DIR_PREFIX).toFile().getAbsoluteFile();
```
coverbeck, line 151: deleteOnExit, see other comment, unless there's a reason to keep it around.

dockstore-client/src/test/java/io/dockstore/client/cli/WesFileIT.java
```diff
@@ -0,0 +1,90 @@
...
+import io.dropwizard.testing.ResourceHelpers;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class WesFileIT {
```
coverbeck, line 13: You should remove `IT` -- makes it look like an integration test, but it's unit tests. I think the ending `IT` has implications in how tests are run on CircleCI (I forget the details), although it may not matter here, but in any case you should remove `IT`.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Thanks, I thought temporary directories would cleanup on exit.

Added: https://github.com/dockstore/dockstore-cli/pull/113/commits/73b5a0bbeb3d04bf035242170002d24af323df96

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
Richard-Hansen: Added: https://github.com/dockstore/dockstore-cli/pull/113/commits/73b5a0bbeb3d04bf035242170002d24af323df96

dockstore-client/src/test/java/io/dockstore/client/cli/WesFileIT.java
Richard-Hansen: Fixed: https://github.com/dockstore/dockstore-cli/pull/113/commits/33d49f0d5644fc8602848115643cd8aa08222d72

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -1135,10 +1135,14 @@ private DockstoreTool getDockstoreTool(String path) {
...
      *
+     * @param entry The path to the desired entry (i.e. github.com/myrepo/myworfklow:version1
+     * @param inlineWorkflow Indicates that the workflow files will be inlined directly into the WES HTTP request
+     * @param paramsPath Path to the parameter JSON file
+     * @param filePaths Paths to any other required files for the WES execution
      */
-    void wesLaunch(String entry, String paramsPath, List<String> filePaths) {
+    void wesLaunch(String entry, boolean inlineWorkflow, String paramsPath, List<String> filePaths) {
         // Only supports workflows for the moment
         throw new RuntimeException("WES launch does not currently support tools. Please launch a workflow instead.");
```
denis-yuen, line 1147: Could be more specific with https://docs.oracle.com/javase/7/docs/api/java/lang/UnsupportedOperationException.html

denis-yuen (approved):

coverbeck (approved):

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
Richard-Hansen: Good point, fixed: https://github.com/dockstore/dockstore-cli/pull/113/commits/9fee96fefb228ee562a24fc3ac2daf076cc2bb11

Richard-Hansen:
> To me, it feels like a fundamental issue with `--attach <file>` is that we can't infer the workflow path from the local path `<file>` in many situations. Somewhere down the line, would it make sense to modify `--attach` to allow the optional specification of the workflow path? For example `--attach <local-path>:<workflow-path>` or something like that?

One of the challenges with `--attach <local-path>:<workflow-path>` is if multiple files reference a single file from multiple different relative paths. In that case, I guess you could use `--attach <local-path>:<workflow-path>` for each `<workflow-path>`, which would solve the issue (assuming no clashing `<workflow-path>` names are uploaded).

It seems the core of the issue is a combination of how WES files are sent, and how WES files are received. I'm currently looking into setting up Toil for testing, and the Toil WES server may not have the same issue as DNAStack or AGC. For example, when uploading a file, part of the HTTP request container some information about the file:
`form-data; filename="nested/Dockstore.wdl"; size=202; name="workflow_attachment"`

From what I'm able to tell, there are two ways the above `filename` has been interpreted:
1. A WES server may interpret this name as a pure filename. There is no directory structure and they will maintain a file object with the name `nested/Dockstore.wdl`, and so all references to that file must be exactly `nested/Dockstore.wdl`. Effectively, whenever you reference a file, they will scan the attached files for a matching `filename`.
2. Toil provisions these files into a directory structure, so `nested/Dockstore.wdl` will be placed in the directory `nested/` (relative to the root of the workflow execution directory), and then all relative references will be made as normal. This seems like an ideal approach.

Depending on how the Toil implementation works out, if its file provisioning functions as expected, it might be worthwhile reaching out to other WES implementers to encourage using directory structures for attached files.

svonworl (approved):

Richard-Hansen:
> I assume workflow input files are located where the workflow is being executed and we will not be provisioning the input JSON or any files mentioned in the input JSON/yml correct?

Any files that are part of the Dockstore workflow are downloaded (this is the default functionality of the download function) so `dockstore.yml` and test parameter files are downloaded. But, the test parameter file is not automatically attached to the request, so the user still specifies their own.

PR #114, opened 2022-01-27
Apptool launch

garyluu:
For start of SEAB-3732

Had a really tough time changing things back and forth trying to figure out what to do.

Current plan is to just have the apptool run in workflow mode while specifically mentioning the oddity in the launch tab.

Here's the very start of the plan. I didn't create a new class or refactor any existing code. It's kind of a mess right now and not entirely sure if this is the way to proceed.

https://docs.google.com/document/d/12q1FH_fbFd9ZqL09PaRuqwJUptghfFgOBcQH9AE6ngA/edit?usp=sharing

Another option is the create an entry mode (./dockstore entry launch) that would be more ambitious and would require a lot of rework

Only the CWL launch with commands have been tested with this PR. will proceed with everything else after approval

TODO: create another endpoint on the webservice so I don't have to run through up to 4 endpoints to get the workflow/apptool

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import javax.ws.rs.core.GenericType;
@@ -102,6 +103,8 @@ public class WorkflowClient extends AbstractEntryClient<Workflow> {
     private final JCommander jCommander;
     private final CommandLaunch commandLaunch;
 
+    private boolean isAppTool;
+
     public WorkflowClient(WorkflowsApi workflowApi, UsersApi usersApi, Client client, boolean isAdmin) {
         this.workflowsApi = workflowApi;
         this.usersApi = usersApi;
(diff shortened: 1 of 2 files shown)


coverbeck:
Can entry names ever clash? I know Natalie added code to prevent app tools from clashing with workflows, are there any other possibilities?

My order of preference, ignoring implementation difficulty:

1. dockstore entry - assuming there are no name clashes
2. dockstore tool - from the end user perspective, it's a tool, even though it's internally implemented it as a workflow.

garyluu:
> Can entry names ever clash? I know Natalie added code to prevent app tools from clashing with workflows, are there any other possibilities?
> 
> My order of preference, ignoring implementation difficulty:
> 
> 1. dockstore entry - assuming there are no name clashes
> 2. dockstore tool - from the end user perspective, it's a tool, even though it's internally implemented it as a workflow.

1. I don't think the entry names can clash. and if it can, we should probably find a way to prevent that.
2. Ya, it is odd that we make the distinction that something is a tool or workflow. The docs link mentions the difficulties of trying to force a workflow into the toolclient. I suspect reworking it would be almost on par with the entry command.

NatalieEO:
> Can entry names ever clash? I know Natalie added code to prevent app tools from clashing with workflows, are there any other possibilities?
> 
Does the CLI do services? The entry name right now could clash with services, but the same work I did to prevent name clashing between workflows and tools should be easy to add to services as well. Tools as we have them now in production should never clash since they would never start with "github.com"

garyluu:
> 

The CLI makes no mention of services so probably not

NatalieEO:
Should we have a broader discussion about whether we want to eventually get rid of tools? Or have some other distinction...eg distinguish types by whether it's docker image based or descriptor file based.

garyluu:
> Should we have a broader discussion about whether we want to eventually get rid of tools? Or have some other distinction...eg distinguish types by whether it's docker image based or descriptor file based. That should make the UI a lot less messy than what it is now...

Should probably discuss further. But to me, nothing is _really_ based on Docker image. It's supposed to be based on the descriptor file. The presence of a specified Docker image only makes it more problematic.

coverbeck (commented):
Was here, waiting for next iteration based on discussion

PR #116, opened 2022-02-01
Improved WES exception handling

Richard-Hansen:
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-3487

This PR introduces some additional exception handling for the WES functionality. Based on the WES API calls, the list of status codes (https://ga4gh.github.io/workflow-execution-service-schemas/docs/#tag/WorkflowExecutionService) we can expect to receive from a WES server includes: 

1. OK: `200`
2. Malformed: `400`
3. Request is unauthorized: `401`
4. Requester is unauthorized: `403`
5. Target not found: `404`
6. Internal server error: `500`

I found that between DNAStack and AGC WES servers, some of the error messages were either in odd formats or not verbose (For example, in the case of a `403`, DNAStack returns the message `error`, and AGC returns the JSON string `{"message":"The security token included in the request is expired"}`). To make sure our error messages are descriptive, I search for each request status code and print out a default descriptive message along with the message provided by the server.

Also, it looks like Jersey throws a `ProcessingException` if the WES URL can't be resolved, so we catch that as well to give a descriptive error message.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
@@ -1,7 +1,6 @@
 package io.dockstore.client.cli.nested;
 
 import java.io.File;
-import java.net.UnknownHostException;
 import java.text.DateFormat;
 import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
@@ -244,7 +244,7 @@ public class ApiClientExtended extends ApiClient {
             // This could be caused by a failed Jersey Interceptor/filter, missing message body writers, or other IO exceptions.
             // Mainly, this error is thrown when the provided WES URL is invalid. For more details, see:
             // https://docs.oracle.com/javaee/7/api/index.html?javax/ws/rs/ProcessingException.html
-            errorMessage(MessageFormat.format("There was an error processing the HTTP request. This could be caused by an invalid URL: {0}",
+            errorMessage(MessageFormat.format("There was an error processing the HTTP request: {0}",
                 ex.getMessage()), Client.CONNECTION_ERROR);
             return null;
         } catch (ApiException ex) {
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
@@ -264,7 +264,7 @@ public class ApiClientExtended extends ApiClient {
                     ex.getCode(), ex.getMessage()), Client.API_ERROR);
                 break;
             case HttpStatus.SC_NOT_FOUND:
-                errorMessage(MessageFormat.format("[{0}] The WES server was unable to locate the target entry: {1}",
+                errorMessage(MessageFormat.format("[{0}] The WES server was unable to locate an entity: {1}",
                     ex.getCode(), ex.getMessage()), Client.API_ERROR);
                 break;
             case HttpStatus.SC_INTERNAL_SERVER_ERROR:
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
@@ -252,28 +252,28 @@ public class ApiClientExtended extends ApiClient {
             // message and the message provided from the WES server in the printed error.
             switch (ex.getCode()) {
             case HttpStatus.SC_BAD_REQUEST:
-                errorMessage(MessageFormat.format("The WES request was malformed: {0}",
-                    ex.getMessage()), Client.API_ERROR);
+                errorMessage(MessageFormat.format("[{0}] The WES request was malformed: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
                 break;
             case HttpStatus.SC_UNAUTHORIZED:
-                errorMessage(MessageFormat.format("The WES request is unauthorized to be performed on the target WES server: {0}",
-                    ex.getMessage()), Client.API_ERROR);
+                errorMessage(MessageFormat.format("[{0}] The WES request is unauthorized to be performed on the target WES server: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
                 break;
             case HttpStatus.SC_FORBIDDEN:
-                errorMessage(MessageFormat.format("The provided credentials are not authorized to make this WES request: {0}",
-                    ex.getMessage()), Client.API_ERROR);
+                errorMessage(MessageFormat.format("[{0}] The provided credentials are not authorized to make this WES request: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
                 break;
             case HttpStatus.SC_NOT_FOUND:
-                errorMessage(MessageFormat.format("The WES server was unable to locate the target entry: {0}",
-                    ex.getMessage()), Client.API_ERROR);
+                errorMessage(MessageFormat.format("[{0}] The WES server was unable to locate the target entry: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
                 break;
             case HttpStatus.SC_INTERNAL_SERVER_ERROR:
-                errorMessage(MessageFormat.format("There was an internal server error processing this WES request: {0}",
-                    ex.getMessage()), Client.API_ERROR);
+                errorMessage(MessageFormat.format("[{0}] There was an internal server error processing this WES request: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
                 break;
             default:
-                errorMessage(MessageFormat.format("There was an unknown error processing this WES request: {0}",
-                    ex.getMessage()), Client.API_ERROR);
+                errorMessage(MessageFormat.format("[{0}] There was an unknown error processing this WES request: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
             }
 
             return null;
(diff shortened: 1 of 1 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -236,6 +240,43 @@ public void createBodyPart(MultiPart multiPart, String key, Object formObject) {
...
+            case HttpStatus.SC_UNAUTHORIZED:
+                errorMessage(MessageFormat.format("[{0}] The WES request is unauthorized to be performed on the target WES server: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
+                break;
+            case HttpStatus.SC_FORBIDDEN:
+                errorMessage(MessageFormat.format("[{0}] The provided credentials are not authorized to make this WES request: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
+                break;
+            case HttpStatus.SC_NOT_FOUND:
+                errorMessage(MessageFormat.format("[{0}] The WES server was unable to locate the target entry: {1}",
```
coverbeck, line 267: Hmm, this could be confusing -- if the CLI invokes an endpoint at `https://<wes-server>/nonexistent/path`, you'll probably get a 404, and the beginning of this message may imply something else.

denis-yuen (approved):
Interesting, Codacy is still working in this repo

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -236,6 +240,43 @@ public void createBodyPart(MultiPart multiPart, String key, Object formObject) {
                     buildResponseHeaders(response),
                     respBody);
             }
+        } catch (ProcessingException ex) {
+            // This could be caused by a failed Jersey Interceptor/filter, missing message body writers, or other IO exceptions.
+            // Mainly, this error is thrown when the provided WES URL is invalid. For more details, see:
+            // https://docs.oracle.com/javaee/7/api/index.html?javax/ws/rs/ProcessingException.html
+            errorMessage(MessageFormat.format("There was an error processing the HTTP request. This could be caused by an invalid URL: {0}",
```
coverbeck, line 247: I don't like "This could be caused by an invalid URL". Presumably there are a ton of reasons a 400 might occur (bad descriptor file, invalid or non-accessible data, unsupported language, etc.).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -236,6 +240,43 @@ public void createBodyPart(MultiPart multiPart, String key, Object formObject) {
...
+            return null;
+        } catch (ApiException ex) {
+            // Different WES servers provide error messages with different levels of verbosity/usefulness, so include both a default
+            // message and the message provided from the WES server in the printed error.
+            switch (ex.getCode()) {
+            case HttpStatus.SC_BAD_REQUEST:
+                errorMessage(MessageFormat.format("[{0}] The WES request was malformed: {1}",
+                    ex.getCode(), ex.getMessage()), Client.API_ERROR);
+                break;
+            case HttpStatus.SC_UNAUTHORIZED:
```
coverbeck, line 258: 401 is really unauthenticated, and the fact that the the standards call it unauthorized always makes me have to look up what's the difference between 401 and 403. :) 

Concrete action: maybe change the message on line 259, but maybe not.

coverbeck (commented):
I'm a little wary of this PR in general. I tried to flag specific cases where you could have confusing messages inline, but I wonder if it's more trouble than it's worth trying to be helpful. If WES takes off, servers will probably return better and better messages, and we'd sort of be obfuscating them.

I understand why you're doing it, and maybe we should, but I feel it will be hard to maintain.

What do others think?

denis-yuen:
> I'm a little wary of this PR in general. I tried to flag specific cases where you could have confusing messages inline, but I wonder if it's more trouble than it's worth trying to be helpful. If WES takes off, servers will probably return better and better messages, and we'd sort of be obfuscating them.
> 
> I understand why you're doing it, and maybe we should, but I feel it will be hard to maintain.
> 
> What do others think?

I buy that. 
Maybe in addition to formatting and inserting the error code, you could also append the original error message

wshands (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
Richard-Hansen: You are correct, the error message changes based on how incorrect the URL is, some exampels:

---
 1.
```
[WES]
url: https://12345.execute-api.us-west-2.amazonaws.com/prod/ga4gh/wes/v1/fake
```
returns a response of:
```
12:28:36.294 [main] ERROR io.dockstore.client.cli.ArgumentUtility - [404] The WES server was unable to locate the target entry: <long based 64 encoded string>
```

---
2.
```
[WES]
url: https://12345.execute-api.us-west-2.amazonaws.com/fake/prod/ga4gh/wes/v1
```
returns a response of:
```
12:34:21.544 [main] ERROR io.dockstore.client.cli.ArgumentUtility - [403] The provided credentials are not authorized to make this WES request: {"message":"Forbidden"}
```

---
3.
```
[WES]
url: notAServerAtAll
```
returns a response of:
```
12:35:32.086 [main] ERROR io.dockstore.client.cli.ArgumentUtility - There was an error processing the HTTP request. This could be caused by an invalid URL: value is null
```

---

So we have a decent variety of errors all related to invalid URLs. (Of the above, only 1 is a 404). The main reason I added the "The WES server was unable to locate target entry" was due to the fact that all 404 responses specified in the WES schema are when a workflow ID is invalid. Perhaps this could be modified to "Invalid WES endpoint"? Not sure, thoughts?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
Richard-Hansen: I added this after the fact when I typed in a horribly wrong URL and got the error message `value is null` from Jersey. For example:
```
[WES]
url: notAServerAtAll
```
returns a response of:
```
12:35:32.086 [main] ERROR io.dockstore.client.cli.ArgumentUtility - There was an error processing the HTTP request. This could be caused by an invalid URL: value is null
```

From what I saw, the `ProcessingException` is thrown by our local client if there is an error with the HTTP request before it's sent. This means that formatting with the HTTP request itself, such as a bad URL or broken headers should be caught here, but things like a bad descriptor file, bad data, etc. wont be (all responses from the WES server are made into `ApiExceptions`, so if there is a bad descriptor file, that will be caught in the following lines).

I'm not attached to `This could be caused by an invalid URL`, so removed it for now: https://github.com/dockstore/dockstore-cli/pull/116/commits/6a36514bbb030011f56973eb3adbb67c59119ce1. Perhaps error hints like this will be worthwhile including in our user documentation once we have that written up.

Richard-Hansen:
> > I'm a little wary of this PR in general. I tried to flag specific cases where you could have confusing messages inline, but I wonder if it's more trouble than it's worth trying to be helpful. If WES takes off, servers will probably return better and better messages, and we'd sort of be obfuscating them.
> > I understand why you're doing it, and maybe we should, but I feel it will be hard to maintain.
> > What do others think?
> 
> I buy that. Maybe in addition to formatting and inserting the error code, you could also append the original error message

The status code and original error message are included in all `ApiException` responses. The general format looks like:
```
[<statusCode>] <Helper message from Dockstore>: <Actual error message from WES server>
```

If we think the `<Helper message from Dockstore>` sections will cause more confusion than anything, I can remove them. Initially, my hope was to not have it in the first place but due to oddly formatted / obscure error messages between WES servers, it seems like error hints from Dockstore could be helpful.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
Richard-Hansen: I'd suggest keeping as-is for now, just to keep our responses with the WES schema: `401 The request is unauthorized.`.

Feel free to push back if there is any disagreement.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
coverbeck: What is a "target entry"? I see it in the AGC message, but I have no idea what it means. I was first parsing it with Dockstore eyes, and thinking "entry" meant tool/workflow/descriptor, but I don't think that's it.

You could use the WES standard response message, "The requested workflow run wasn't found", but that would only apply in some cases.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
```diff
@@ -236,6 +240,43 @@ public void createBodyPart(MultiPart multiPart, String key, Object formObject) {
...
+                errorMessage(MessageFormat.format("[{0}] The WES request is unauthorized to be performed on the target WES server: {1}",
```
coverbeck, line 259: Instead of "the target WES server", you could show the actual server or url. For example, if you had a typo in the url in the ~/.dockstore/config, showing the url could be a helpful debugging tip. Or is this already displayed via `ex.getMessage()`?

Optional or a future ticket, especially given my reservations about this approach.

coverbeck (approved):
I still have my doubts about this; there's the risk of displaying duplicative, wrong, and/or contradictory messages.

But it's fairly minor and we can always adjust it later.

garyluu (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ApiClientExtended.java
Richard-Hansen: We don't surface the URL. I think that a broader ticket for a WES debug mode would be valuable, especially for dealing with WES-server specific issues. Created a followup ticket: https://github.com/dockstore/dockstore/issues/4709.

PR #118, opened 2022-02-02
Postgres update

denis-yuen:
Hotfix equivalent of https://github.com/dockstore/dockstore-cli/pull/117


denis-yuen:
Think snyk is just being weird, singularity pip should be unrelated

coverbeck (approved):
Is this just in case we do a hotfix, and/or does it make `hf finish` easier?

denis-yuen:
"just in case we do a hotfix"  I think. 
It looks like develop has this stuff already

Richard-Hansen (approved):

PR #119, opened 2022-02-07
Don't throw error message if Jar already present

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-3147
Upgrading CLI from 1.10.0 to 1.11.0 works but throws error message
Return the stdout and stderr so you can see message from dockstore script
Don't report a failure just because the JAR is already there.

### DIFF
--- a/dockstore-client/bin/dockstore
+++ b/dockstore-client/bin/dockstore
@@ -157,7 +157,7 @@ if [ "$1" = "self-install" ]; then
     if [ -r "$DOCKSTORE_JAR" ]; then
         echo "The self-install jar already exists at $DOCKSTORE_JAR."
         echo "If you wish to re-download, delete it and rerun \"$0 self-install\"."
-        exit 1
+        exit 0
     fi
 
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):

wshands:
The dockstore script is downloaded automatically.

I did see this in the Intellij console when debugging:
```
Connected to the target VM, address: '127.0.0.1:52268', transport: 'socket'
Java HotSpot(TM) 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
Current Dockstore version: 1.12.0-alpha.0-SNAPSHOT
Upgrading to most recent stable release (1.12.0-alpha.0-SNAPSHOT -> 1.11.2)
The self-install jar already exists at /Users/waltershands/.dockstore/self-installs/dockstore-client-1.11.2.jar.
If you wish to re-download, delete it and rerun "/Users/waltershands/bin/dockstore self-install".
```

svonworl (approved):

Richard-Hansen (approved):

PR #121, opened 2022-02-09
Feature/3785/testing wes with toil

Richard-Hansen:
This PR adds:
1. A Toil WES server (launched with `docker compose`) to CI. This can also be run locally by running `docker compose up` from the `toil/` directory of the CLI repo.
2. Some basic WES tests launching workflows. See the CircleCi `wes-toil-test` job.
3. A new test repo https://github.com/dockstore-testing/wes-testing, which includes published WDL workflows used in the live tests

Each workflow is given 100 seconds to reach the desired state (for example, `COMPLETED`) before the test fails. This has the potential to be flakey, but workflows are currently running for about 30 seconds, so there is a decent buffer. We can up this if needed.

Currently, the Dockstore CLI doesn't print easily parseable values for any WES request, so there is some patchwork parsing used as part of the testing process. As part of https://github.com/dockstore/dockstore/issues/4709, we should see the CLI printing better values by default (and only the verbose logs when `--debug` is used).

There are some bugs related to the CLI + Toil integration currently:
1. Toil does not support workflow imports.
2. The Dockstore CLI fails to process the `serviceInfo` endpoint response (It looks to be an issue with decoding the JSON), created a followup issue: https://github.com/dockstore/dockstore/issues/4716

---


dockstore-client/src/test/java/io/dockstore/client/cli/WesToilIT.java
```diff
@@ -0,0 +1,155 @@
...
+            "--json", ResourceHelpers.resourceFilePath("wesIt/w1_test.json"),
+            "--inline-workflow"
+        };
+        Client.main(commandStatementRun);
+        assertTrue("A helper command should be printed to stdout when a workflow is successfully started", systemOutRule.getLog().contains("dockstore workflow wes status --id"));
+
+        final String runId = findWorkflowId(systemOutRule.getLog());
+        final boolean isSuccessful = waitForWorkflowState(runId, COMPLETED_STATE);
+
+        if (!isSuccessful) {
```
coverbeck, line 86: I'm sure you know this, and if you prefer this style, that's fine, but just in case, you could rewrite the block as one line:

`assertTrue("The workflow did not succeed in time.", isSuccessful);`

dockstore-client/src/test/java/io/dockstore/client/cli/WesToilIT.java
```diff
@@ -0,0 +1,155 @@
...
+        if (!isError) {
+            fail("The workflow was not supposed to succeed");
+        }
+    }
+
+    @Test
+    public void testBasicLaunch3() throws InterruptedException {
+        String[] commandStatementRun = new String[]{ "workflow", "wes", "launch",
+            "--config", TOIL_CONFIG,
+            "--entry", "github.com/dockstore-testing/wes-testing/single-descriptor-no-input:main",
```
coverbeck, line 115: You could have a tag if you think main will ever change.

coverbeck (approved):
This is great!

dockstore-client/src/test/java/io/dockstore/client/cli/WesToilIT.java
Richard-Hansen: Yup, thanks for pointing this out, I guess I wrote the first one weird and all subsequent tests inherited that format.

Fixed: https://github.com/dockstore/dockstore-cli/pull/121/commits/c58211042518e6b5fd9d64f15f9e636922d73d9b

dockstore-client/src/test/java/io/dockstore/client/cli/WesToilIT.java
Richard-Hansen: Ideally, I think main shouldn't change to help with test maintenance. Added a note to the test repo's README to that effect: https://github.com/dockstore-testing/wes-testing/blob/main/README.md

Feel free to push back if there is any disagreement.

denis-yuen (commented):
clearing notification, think you're still working through build issues

PR #124, opened 2022-02-22
Feature/4697/wes directory attach

Richard-Hansen:
This PR adds functionality to the Dockstore CLI to allow users to attach entire directories to their WES requests. This is overloaded in the normal attach parameters `--attach`/`-a`. The Dockstore CLI will walk the directory tree of any passed directories and add all files found. These files will be named relative to the directory root passed in.

---

For example, if the user specifies the directory `/path/to/dir` with files `/path/to/dir/file1` and `/path/to/dir/nested/file2`, the files attached to the WES HTTP request will be named `file1` and `nested/file2`. If the user specifies a relative path to a directory, i.e. `to/dir`, this path will be converted to the absolute path `/path/to/dir`, and files will be added as shown above.

---

To test file attachments, more Toil WES tests were added, alongside two more test repos:
1. https://github.com/dockstore-testing/wes-testing/tree/main/workflows/wdl/w4
2. https://github.com/dockstore-testing/wes-testing/tree/main/workflows/wdl/w5

Just to clarify how file attachments are handled by the Dockstore CLI:
1. Absolute paths to a single file -> Throw an error. We don't know how to name the file in the WES request.
2. Relative paths to a single file -> Attach with the name equal to the provided relative path.
3. Absolute path to a directory -> Upload all files in the directory tree named relative to the directory itself.
4. Relative path to a directory -> Convert to an absolute path and follow the same pattern as number 3.
5. Path expansions (e.g. `~`) -> Not handled.

Although the workflows reach the `COMPLETE` state in Toil, their outputs are incorrect, as it seems the provisioned files aren't referenced correctly.

- [] Check that you pass the basic style checks and unit tests by running `mvn clean install`


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesFile.java
```diff
@@ -51,6 +52,11 @@ public String getName() {
         // If the file was provisioned locally from an entry on Dockstore, we can remove the path to the temporary directory
         // from the file's absolute path to obtain a relative path.
         if (this.removablePrefix != null) {
+
+            if (!(Path.of(removablePrefix)).isAbsolute()) {
+                errorMessage("The provided removable prefix must be absolute", CLIENT_ERROR);
```
coverbeck, line 57: Is a end-user going to understand this message? What is a removable prefix? I sort of get it from the comment on line 52, but an end-user won't see that.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesLauncher.java
```diff
@@ -233,13 +233,13 @@ public static String combineTrsUrlComponents(WorkflowClient workflowClient, Stri
...
         }
 
-        // Verify the file path exists
-        // 1. If the path is to a directory, return an empty optional
+        // If the path isn't a file or a directory, we've done something wrong on our end.
+        // We don't want to attach a directory directly, so return an empty optional instead
         Path path = Paths.get(filePath);
         if (Files.isDirectory(path)) {
             return Optional.empty();
         } else if (!Files.isRegularFile(path)) {
```
coverbeck, line 241: Will this fail on symbolic links? If so, do we want that?

If we're talking about uploading genomic data (are we?), then I could see how you might use them.

If I'm right, can be a follow up ticket.

Richard-Hansen: Good point, I hadn't considered that. 

Taking a look at the Java Docs (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#isRegularFile(java.nio.file.Path,java.nio.file.LinkOption...), symbolic links are followed by default, and should be read properly:

> By default, symbolic links are followed and the file attribute of the final target of the link is read.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesFile.java
Richard-Hansen: Fixed: https://github.com/dockstore/dockstore-cli/pull/124/commits/5b1389e7f400aa26dde0c1c70dc273386b02cdaa

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesFile.java
```diff
@@ -51,6 +52,11 @@ public String getName() {
         // If the file was provisioned locally from an entry on Dockstore, we can remove the path to the temporary directory
         // from the file's absolute path to obtain a relative path.
         if (this.removablePrefix != null) {
+
+            if (!(Path.of(removablePrefix)).isAbsolute()) {
+                errorMessage(MessageFormat.format("Unable to calculate relative name for HTTP file attachment: {0}", super.getName()), CLIENT_ERROR);
```
coverbeck, line 57: I'm still not clear what I would do if I saw this message. As an end-user, what did I do wrong, and how can I fix it?

coverbeck (approved):
There is a ticket to document the WES work for 1.12, right? We should make sure to document this and all the other features you've added.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WesFile.java
Richard-Hansen: I don't think this will ever be surfaced to the user unless we internally use the class incorrectly, so it's more of a logical sanity check. That being said, if this error does appear, the user's best bet is to change the way they are specifying the attachment file.

Perhaps theres a better error type for this? It's more of an `INTERNAL_ERROR` (doesn't exist in the CLI) rather than a `CLIENT_ERROR`. I could also just throw an exception...

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/WesToilIT.java
```diff
@@ -150,6 +150,107 @@ public void testCancel() throws InterruptedException {
...
+    }
+
+    @Test
+    public void testDirectoryAttachment() throws InterruptedException {
+        // These tests pass, and the files provisioned by Toil look correct, but the outputs are not.
+        String[] commandStatementRun = new String[]{ "workflow", "wes", "launch",
+            "--config", TOIL_CONFIG,
+            "--entry", "github.com/dockstore-testing/wes-testing/single-descriptor-nested-input:main",
+            "--json", ResourceHelpers.resourceFilePath("wesIt/w4_1_test.json"),
+            "-a", ResourceHelpers.resourceFilePath("wesIt/w4_nested"),
```
denis-yuen, line 162: use `--attach` in one of these tests just to be thorough

denis-yuen (approved):
with above comment

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/WesToilIT.java
Richard-Hansen: Added: https://github.com/dockstore/dockstore-cli/pull/124/commits/4672000da6be9146b15fba12a33e9be4d16fa5c4

wshands (approved):

PR #125, opened 2022-02-23
WES response JSON printing

Richard-Hansen:
This PR changes the print formatting of WES responses. We are now printing the responses as pure JSON (rather than Stringified Objects). So the user will see:
```
{
  "workflow_type_versions" : {
    "WDL" : {
      "workflow_type_version" : [ "1.0", "draft-2" ]
    }
  },
  "supported_wes_versions" : [ "1.0.0" ],
  "supported_filesystem_protocols" : null,
  "workflow_engine_versions" : null,
  "default_workflow_engine_parameters" : null,
  "system_state_counts" : null,
  "auth_instructions_url" : null,
  "contact_info_url" : null,
  "tags" : {
    "cromwell_service_health" : "True",
    "description" : "WES adapter for Cromwell workflow engine service.",
    "name" : "remote_cromwell_wes_adapter",
    "updated_at" : "2022-02-23T17:37:54.672500Z"
  }
}
```

instead of:
```
class ServiceInfo {
    workflowTypeVersions: {WDL=class WorkflowTypeVersion {
        workflowTypeVersion: [1.0, draft-2]
    }}
    supportedWesVersions: [1.0.0]
    supportedFilesystemProtocols: null
    workflowEngineVersions: null
    defaultWorkflowEngineParameters: null
    systemStateCounts: null
    authInstructionsUrl: null
    contactInfoUrl: null
    tags: {cromwell_service_health=True, description=WES adapter for Cromwell workflow engine service., name=remote_cromwell_wes_adapter, updated_at=2022-02-23T17:38:53.388191Z}
}
```

The two WES POSTs (RunWorkflow and CancelRun) do not use the JSON printing, as the response Object in JSON looks like: `{"runId": "123-456-789"}`, so I left it as just the Run ID `123-456-789`.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -43,6 +43,8 @@ import com.amazonaws.auth.profile.ProfilesConfigFile;
 import com.amazonaws.regions.AwsProfileRegionProvider;
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.ParameterException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.command.InfoCmd;
 import com.github.dockerjava.core.DefaultDockerClientConfig;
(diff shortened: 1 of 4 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1115,9 +1120,12 @@ private void wesStatus(WorkflowExecutionServiceApi clientWorkflowExecutionServic
     private void wesRunLogs(WorkflowExecutionServiceApi clientWorkflowExecutionServiceApi, final String workflowId) {
         try {
             RunLog response = clientWorkflowExecutionServiceApi.getRunLog(workflowId);
-            out(response.toString());
+            ObjectMapper mapper = new ObjectMapper();
```
coverbeck, line 1123: Just being paranoid, if a WES server were compromised and sent us an evil response, could this lead to code being executed?

I don't think so (it was a YAML parser that had an exploit, which has been patched), but like I said, being paranoid.

Added @david4096 as a reviewer.

coverbeck (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
david4096: Took a look, you can do type overrides in Jackson like this:

```
{
  "@class":"MyApp.Obj",
  val:[
    "java.lang.Long",
    1
  ]
}
```

However, it appears the exploit requires you explicitly turn on `mapper.enableDefaultTyping`. I think we are safe based on my current understanding.

https://adamcaudill.com/2017/10/04/exploiting-jackson-rce-cve-2017-7525/

david4096 (approved):

denis-yuen (approved):

PR #126, opened 2022-03-03
Feature/dock 2057/improve launch failure message

svonworl:
This PR changes the dockstore CLI to print information about the exception thrown upon a launch failure (the type of exception, message [if present], and a stacktrace), in addition to the error message.  It also includes a cleanup of the adjacent launch code. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2057
dockstore/dockstore#4713


coverbeck (approved):

denis-yuen (approved):

Richard-Hansen (approved):
Looks good.

At some point in the future (not for this ticket) we should probably move all parameter parsing logic into jCommander classes (and use jCommander for all parameter reading as well). Hopefully that'll help with CLI consistency.

PR #130, opened 2022-03-08
Move WES out of beta

Richard-Hansen:
WES support in the CLI is tested with WDL, so changed the note to that effect. I also changed the all-caps text because it seemed excessively 'loud', but can change it back if that's the recommended format for notes.

This is also seems like a good PR to finish off the develop changes and cut the release branch.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -1653,7 +1653,7 @@ public abstract class AbstractEntryClient<T> {
         out("Global Optional Parameters:");
         out("  --wes-url <WES URL>                 URL where the WES request should be sent, e.g. 'http://localhost:8080/ga4gh/wes/v1'");
         out("");
-        out("NOTE: WES SUPPORT IS IN BETA AT THIS TIME. RESULTS MAY BE UNPREDICTABLE.");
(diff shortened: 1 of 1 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1653,7 +1653,7 @@ private void printWesHelpFooter() {
         out("Global Optional Parameters:");
         out("  --wes-url <WES URL>                 URL where the WES request should be sent, e.g. 'http://localhost:8080/ga4gh/wes/v1'");
         out("");
-        out("NOTE: WES SUPPORT IS IN BETA AT THIS TIME. RESULTS MAY BE UNPREDICTABLE.");
+        out("NOTE: Currently only WDL workflows are supported for WES requests. Further language support is under development.");
```
coverbeck, line 1656: If AGC magically added support for CWL tomorrow, in theory the CLI would just work, wouldn't it?

That said, I think this is fine.

coverbeck (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Richard-Hansen: It should, it just hasn't been tested.

PR #155, opened 2022-05-18
Make test parameter files optional (#4763)

amarjandu:
https://ucsc-cgl.atlassian.net/browse/DOCK-2087

dockstore/dockstore#4763


coverbeck (changes requested):
1. Can you reference the issue this is fixing? I eventually found it, but I went looking for issue 2087 to start with because that's the name of your branch. Maybe I should have recognized the 4763 in the title right away, but it's not a clickable link, which makes things easier. What is 2087, BTW?
2. As a minor nit, it's not crucial, but it would be nice if you followed the HubFlow convention of branch naming. You don't have to use HubFlow to create the branch if you don't want, but the convention is `feature/\<issue>/\<optional description>
3. Could you please add some sort of test(s) for this?

coverbeck:
> As a minor nit, it's not crucial, but it would be nice if you followed the HubFlow convention of branch naming. You don't have to use HubFlow to create the branch if you don't want, but the convention is `feature/<issue>/<optional description>

To be clear, it's not worth changing the branch name for this PR, but going forward it would be good to follow the pattern.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -466,16 +466,19 @@ public void launchWithArgs(final String entry, final String localEntry, final St
                         final String descriptor = descriptorType.getValue().toLowerCase();
                         LanguageClientInterface languageClientInterface = convertCLIStringToEnum(descriptor);
                         DescriptorLanguage language = DescriptorLanguage.convertShortStringToEnum(descriptor);
-
+                        if (workflow.getDefaultTestParameterFilePath() == null) {
```
Richard-Hansen, line 469: If I understand this block correctly, if there isn't a default test parameter file, and one wasn't passed in via the CLI, an error is thrown. This means that if a user has a workflow that doesn't take any inputs, they'll still need to pass in a dummy `--json` or `--yaml` parameter, correct?

For example, launching something like https://dockstore.org/workflows/github.com/dockstore-testing/dockstore-whalesay2:master?tab=info, should be launch-able without any default or explicit parameter files:
Screen Shot 2022-05-20 at 10 28 09 AM

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -466,16 +466,19 @@ public void launchWithArgs(final String entry, final String localEntry, final St
...
+                            conditionalErrorMessage((yamlRun != null) == (jsonRun != null),
```
Richard-Hansen, line 470: If a user passes in values under both `--yaml` and `--json`, this will trigger. It's a niche case, but the error message might still be confusing. But, depending on the comment above, this might not be needed at all.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -466,16 +466,19 @@ public void launchWithArgs(final String entry, final String localEntry, final St
...
                         try {
                             switch (language) {
                             case CWL:
-                                conditionalErrorMessage((yamlRun != null) == (jsonRun != null), "Either --json or --yaml is required", CLIENT_ERROR);
                                 languageClientInterface.launch(entry, false, yamlRun, jsonRun, null, uuid);
                                 break;
                             case WDL:
                             case NEXTFLOW:
-                                conditionalErrorMessage(jsonRun == null, "dockstore: missing required flag --json", CLIENT_ERROR);
+                                conditionalErrorMessage((yamlRun != null), "Nextflow only supports --json, --yaml cannot be used", CLIENT_ERROR);
```
svonworl, line 481: The `case` statements fall through, so this code is executed for both the `WDL` and `NEXTFLOW` cases.

amarjandu:
> 1. Can you reference the issue this is fixing? I eventually found it, but I went looking for issue 2087 to start with because that's the name of your branch. Maybe I should have recognized the 4763 in the title right away, but it's not a clickable link, which makes things easier. What is 2087, BTW?

DOCK-2087 was the Jira issue, I switched to the GitHub issue after our conversation about the benefits of using the public facing GitHub issue.

I added links to both issues at the PR description. 

> 2. As a minor nit, it's not crucial, but it would be nice if you followed the HubFlow convention of branch naming. You don't have to use HubFlow to create the branch if you don't want, but the convention is `feature/<issue>/<optional description>

Sure not a problem, are there a document that has best-practice guidelines that I can refer to? 

> 3. Could you please add some sort of test(s) for this? 

Will do.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
amarjandu: @Richard-Hansen 
Yeah the idea was that if a workflow does not specify a default test parameter file, the user should specify one.

Is there an attribute to a workflow that can tell me if there needs to be a test input file? That would probably be the best way to decide if parameter is required.

amarjandu: @coverbeck @denis-yuen do you think its worth inspecting if there is a default test parameter files specified in the workflow, I was trying to avoid users running into an error if there was no default specified but was required. If this is a non-issue I'd prefer to remove the check.

coverbeck: TLDR; I think don't bother.

You wouldn't use `Workflow.getDefaultTestParameterFilePath` for this. It's a hint for Dockstore indexing the workflow; it tells Dockstore where to look for a test parameter file when creating a new version via refresh. For example,

1. For the repo foo, the user sets the default test parameter file path  to `/goo.json`.
2. The user adds a new branch `bar` in Git
3. The user clicks refresh in Dockstore for the repo foo
4. Dockstore reads the repo foo, looking for the file /goo.json in Git. If it finds it, /goo.json and its contents get added to the version `bar` on Dockstore. If it doesn't find it, then it doesn't get added.

Finally, whether or not `goo.json` gets added to the version on Dockstore not, it does not indicate whether the workflow/tool **requires** inputs or not.

To do that, you would need to do something like what the [convert](https://github.com/dockstore/dockstore-cli/blob/b7cb750e99700885b009a401e99ea98d5dceced5/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java#L620) method does. And even then, I think in theory it could return only optional parameters, so it wouldn't be 100% reliable.

Just let the workflow engine complain when running the workflow, IMHO.

denis-yuen:
> Sure not a problem, are there a document that has best-practice guidelines that I can refer to?

https://wiki.oicr.on.ca/display/DOC/System+Development+Life+Cycle could be useful

denis-yuen (commented):
was here, comments make sense, will wait for next revision

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1010,6 +1011,9 @@ void preValidateLaunchArguments(List<String> args) {
         List<String> argsCopy = new ArrayList<>(args);
         String jsonFile = optVal(argsCopy, "--json", null);
         String yamlFile = optVal(argsCopy, "--yaml", null);
+        conditionalErrorMessage((jsonFile != null) && (yamlFile != null),
+                "Specify a test parameter file with either --json or --yaml not both",
```
coverbeck, line 1015: I'd add a comma after --yaml, i.e., `or --yaml, not both`. 

Also, I realize the message in this context is for when they're trying to specify at least one parameter file, but technically the other option is not to specify a parameter file at all. 

How about:

When specifying a test parameter file, use either --json or --yaml, but not both.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -171,19 +171,22 @@ public long launchPipeline(String entryVal, boolean localEntry, ToolDescriptor.T
...
+                        exceptionMessage(ex, "The workflow entry does not exist. Did you mean to launch a local workflow or a tool?",
+                                ENTRY_NOT_FOUND);
+                    }
+                } catch (Exception ex) {
+                    exceptionMessage(ex, ex.getMessage(), GENERIC_ERROR);
                 }
-            } catch (Exception ex) {
-                exceptionMessage(ex, ex.getMessage(), GENERIC_ERROR);
+            } else {
+                LOG.debug("Input files were not passed, attempting to continue.");
```
coverbeck, line 189: I would just say "No test parameter file provided, skipping provisioning".

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CwltoolLauncher.java
```diff
@@ -47,11 +47,13 @@ public List<String> buildRunCommand() {
 
         extraFlags = extraFlags.stream().map(string -> string.split(",")).flatMap(Arrays::stream).map(this::trimAndPrintInput)
                 .collect(Collectors.toList());
-
         // Create base execution command
         CWLRunnerInterface cwlRunner = CWLRunnerFactory.createCWLRunner();
         command = cwlRunner.getExecutionCommand(workingDirectory + "/outputs/", workingDirectory + "/tmp/", workingDirectory + "/working/",
-                primaryDescriptor.getAbsolutePath(), provisionedParameterFile.getAbsolutePath());
+                primaryDescriptor.getAbsolutePath(), java.util.Optional.empty());
```
coverbeck, line 53: Nit, unless there's a conflict, I'd make this `Optional.empty()` and do an import

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -1393,4 +1394,69 @@ public void cwlNullInputParameter() {
         // run simple echo null tool
         runTool(nullCWL, nullJSON, false);
     }
+
+    @Test
+    public void missingTestParameterFileWDL() {
+        // Run a workflow without specifying test parameter files, deffer errors to the
```
coverbeck, line 1400: defer typo

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -1393,4 +1394,69 @@ public void cwlNullInputParameter() {
         // run simple echo null tool
         runTool(nullCWL, nullJSON, false);
     }
+
+    @Test
+    public void missingTestParameterFileWDL() {
```
coverbeck, line 1399: I forget, does WDL always require a test parameter file?

If not, can you add a test that successfully runs such a WDL. If yes, ignore this comment.

dockstore-client/src/test/resources/no-input-echo.cwl
```diff
@@ -0,0 +1,5 @@
+cwlVersion: v1.0
+class: CommandLineTool
+baseCommand: [echo, hello]
+inputs: []
+outputs: []
```
coverbeck, line 5: GitHub complaining about missing EOL at EOF

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
denis-yuen: "When": seems a little subtle, maybe 
"If specifying a test parameter file, use either --json or --yaml, but not both."?

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -240,7 +241,7 @@ private void yamlAndJsonEntryCorrect(String entryType) {
         args.add("--json");
         args.add(jsonTestParameterFile.getAbsolutePath());
         exit.expectSystemExitWithStatus(CLIENT_ERROR);
-        exit.checkAssertionAfterwards(() -> Assert.assertTrue(systemErrRule.getLog().contains("Missing required flag")));
+        exit.checkAssertionAfterwards(() -> Assert.assertTrue(systemErrRule.getLog().contains("Specify a test parameter file with either --json or --yaml not both")));
```
denis-yuen, line 244: Use a constant string shared with the implementation to make it easier for the next developer that changes the error text. i.e. developer will more easily be able to see that the error message is referenced in a test

denis-yuen (approved):
Minor comment

coverbeck (approved):

PR #161, opened 2022-06-09
Feature/dock 2129/yaml file check

fhembroff:
The main idea behind is to do the following (in this order),

1. Check that the given directory exists
2. Check that `.dockstore.yml` exists
3. Check that the provided yaml is valid (ie. does not have indentation errors)
4. Check that all fields in the yaml have recognizable names
5. Check that the yaml has all required fields 
6. Check that all specified files exist


dockstore-client/generated/src/main/resources/pom.xml
```diff
@@ -342,6 +342,36 @@
       <version>1.3.5</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
```
denis-yuen, line 345: I suspect that not all of these are needed, as discussed, it might be because of the granularity of the code we're re-using (i.e. may need to split out some of the re-usable code)

dockstore-client/pom.xml
```diff
@@ -296,8 +298,33 @@
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
-    </dependencies>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <version>6.1.7.Final</version>
```
denis-yuen, line 304: on a more micro-level, if you did need to add these dependencies, I think you can skip the versions. Our dockstore-bom should cover some of these at least

denis-yuen (commented):
In our discussion, I think you mentioned an indirect connection to `AbstractWorkflowResource` or similar, can you elaborate on that. It isn't obvious from the code that https://github.com/dockstore/dockstore/blob/develop/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java#L3-L37 has dependencies on these new dependencies

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
```diff
@@ -0,0 +1,209 @@
...
+            out(dockstoreYmlPath + " is a valid yaml file");
+        } catch (Exception ex) {
+            throw new ValidateYamlException(ERROR_MESSAGE + dockstoreYmlPath.toString() + " is not a valid yaml file:\n" + ex.getMessage());
+        }
+
+        // Running validate first, as if readAsDockstoreYaml12 is run first and a parameter is incorrect it is difficult to understand
+        // Determine if any of the parameters names are incorrect (does not verify if all the parameters are there)
+        try {
+
+            DockstoreYamlHelper.validateDockstoreYamlProperties(Files.readString(dockstoreYmlPath));
```
coverbeck, line 172: Minor nit, I'd just read the file just once and save the value (you read it again on line 180). And catch that specific exception, e.g., you're unable to read the file.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
```diff
@@ -0,0 +1,209 @@
...
+    private void validateHelp() {
+        printHelpHeader();
+        out("Usage: dockstore " + getEntryType().toLowerCase() + " validate --help");
+        out("       dockstore " + getEntryType().toLowerCase() + " validate [parameters]");
+        out("");
+        out("Description:");
+        out("  Verifies that a .dockstore.yml file is valid and that all required files are present");
+        out("");
+        out("Required Parameters:");
+        out("  --path <path>                                                          Complete entry path in computer (ex. /home/usr/test)");
```
coverbeck, line 78: I think "on computer"

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
```diff
@@ -0,0 +1,209 @@
...
+
+/*
+    GENERAL STEPS:
+    1. Verfiy that YAML syntax is valid
+    2. Verify that all fields are named correctly
+    3. Verify that all required fields are present
+    4. Verify that all files exist
+ */
+
+public class YamlVerify extends WorkflowClient {
```
coverbeck, line 39: I realize it may be tricky the way the code is structured, and I'm fuzzy on those details, but it doesn't seem like you should be extending WorkflowClient.

1. If I wrote a .dockstore.yml that had only tools and/or services, I would run `dockstore workflow validate`, which seems counterintuitive.
2. In the constructor on line 44, I don't think you use any of those parameters (I could be wrong though).

I'd argue for something like `dockstore validate`, or maybe `dockstore yaml validate`, which could give us the option in the future to do something like `dockstore yaml generate`

coverbeck (commented):
Looks really cool. I think there are additional validations you might be able to do, but maybe not, and if so, they could certainly wait.

dockstore-client/generated/src/main/resources/pom.xml
fhembroff: > 
In regards to `AbstractWorkflowResource.java` it appears that some validation of the Dockstore Yaml is being done there.
https://github.com/dockstore/dockstore/blob/3680542eda82ec9728418b6a8237e591936c484e/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L364

Validation that is being done in  `AbstractWorkflowResource.java` includes [verifying that WDL is not being used as a subclass](https://github.com/dockstore/dockstore/blob/3680542eda82ec9728418b6a8237e591936c484e/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L489).

fhembroff:
One of the issues I was having with `DockstoreYamlHelper.readAsDockstoreYaml12(...)` is that it doesn't always fail if DockstoreYaml is incorrect according to the guidelines stated [here](https://docs.dockstore.org/en/stable/assets/templates/workflows/workflows.html).

For example, a workflow `.dockstore.yml` with two workflows should not be allowed if both do not have a name, however in my testing it passes.

I also found that `DockstoreYamlHelper.readAsDockstoreYaml12(...)` is inconsistent at times, for example the following would fail because the subclass on the service is invalid,
```
version: 1.2
workflows:
  -  name: foobar
     subclass: wdl
     authors:
       - name: Denis Yuen
         orcid: 0000-0002-6130-1021
       - name: UCSC Genomics Institute
         role: Institute
         affiliation: UCSC
     publish: true
     primaryDescriptorPath: /Dockstore2.wdl
     testParameterFiles:
       - /dockstore.wdl.json
     filters:
       branches:
         - develop
       tags:
         - gwas*
  -  name: foobar2
     subclass: cwl
     publish: no
     primaryDescriptorPath: /Dockstore.cwl
     testParameterFiles:
       - /dockstore.cwl.json
  # A workflow with no name:
  -  subclass: CWL
     primaryDescriptorPath: /Dockstore.cwl
     testParameterFiles:
       - /dockstore.cwl.json
service:
  subclass: foobar
  name: UCSC Xena Browser
  author: UCSC Genomics Institute
  authors:
    - name: UCSC Genomics Institute
      role: Institute
      affiliation: UCSC
  description: |
    The UCSC Xena browser is an exploration tool for public and private,
    multi-omic and clinical/phenotype data.
    It is recommended that you configure a reverse proxy to handle HTTPS
  publish: True
  files:
    - docker-compose.yml
    - README.md
    - stand_up.sh
    - load.sh
    - port.sh
    - stop.sh
    - xena-all.json

  scripts:
    start: stand_up.sh
    postprovision: load.sh
    port: port.sh
    stop: stop.sh

  environment:
    httpPort:
      default: 7222
      description: The host's HTTP port. The default is 7222.

  data:
    dataset_1:
      targetDirectory: xena/files
      files:
        tsv:
          description: Data for Xena in TSV format
        metadata:
          description: The metadata, in JSON format, that corresponds to the previous tsv 
```
However, this passes,
```
version: 1.2
workflows:
  -  subclass: ERROR_SHOULD_OCCUR
     authors:
       - name: Denis Yuen
         orcid: 0000-0002-6130-1021
       - name: UCSC Genomics Institute
         role: Institute
         affiliation: UCSC
     publish: true
     primaryDescriptorPath: /Dockstore2.wdl
     testParameterFiles:
       - /dockstore.wdl.json
     filters:
       branches:
         - develop
       tags:
         - gwas*
  -  subclass: cwl
     publish: no
     primaryDescriptorPath: /Dockstore.cwl
     testParameterFiles:
       - /dockstore.cwl.json
```
even though the `subclass` is invalid.

I am currently thinking some of these errors should be corrected in  `DockstoreYamlHelper`, Does anyone have any thoughts?

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
```diff
@@ -0,0 +1,209 @@
...
+            if (!Files.exists(pathToFile)) {
+                missingFiles += pathToFile.toString() + " does not exist\n";
+            }
+        }
+        return missingFiles;
+    }
+
+    // Determines if all the files in dockstoreYaml12 exist
+    private static void allFilesExist(DockstoreYaml12 dockstoreYaml12, String basePath) throws ValidateYamlException {
+        String missingFiles = "";
```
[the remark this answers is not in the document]
fhembroff: Thanks for the suggestion, I have fixed this

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
fhembroff: Thanks for the feedback, I have fixed this.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
fhembroff: Fixed :+1:

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
fhembroff: Good Idea. So I should create a class called `YamlClient` (which would implement ` AbstractEntryClient<T>`) and another class `ValidateYaml` which would extend `YamlClient` and in `ValidateYaml` would be all of the code in `YamlVerify`?

dockstore-client/pom.xml
fhembroff: I removed all the version numbers.

dockstore-client/generated/src/main/resources/pom.xml
fhembroff: I removed the following from `pom.xml`
 - `org.hibernate.common:hibernate-commons-annotations`
 - `jakarta.validation:jakarta.validation-api`
 
However, I found that the following are required,
 - `org.hibernate:hibernate-validator`
 - `org.glassfish:jakarta.el`
 
As even though it will compile, and the CLI will work normally, in certain circumstances you will get an error, for example if you execute `java -jar dockstore-client-1.13.0-SNAPSHOT-shaded.jar yaml validate --path /src/test/resources/YamlVerifyTestDirectory/2ToolsWithNoName` without the above required dependencies you will get this error,
```
HV000183: Unable to initialize 'javax.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead
at io.dockstore.client.cli.YamlVerify.dockstoreValidate(YamlVerify.java:211)
at io.dockstore.client.cli.YamlVerify.validateChecker(YamlVerify.java:226)
at io.dockstore.client.cli.YamlVerify.processEntryCommands(YamlVerify.java:103)
at io.dockstore.client.cli.Client.run(Client.java:733)
at io.dockstore.client.cli.Client.main(Client.java:630)
```

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
coverbeck: > Good Idea. So I should create a class called `YamlClient` (which would implement ` AbstractEntryClient<T>`) and another class `ValidateYaml` which would extend `YamlClient` and in `ValidateYaml` would be all of the code in `YamlVerify`?

I don't think you should extend AbstractEntryClient -- an entry is a tool, workflow, or service, and this feature is sort of global.

It looks like there are a [couple of other clients ](https://github.com/dockstore/dockstore-cli/blob/439478e67c1985c6b8306865251e6e58ac465654/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java#L714)that don't extend AbstractEntryClient; perhaps you could follow that pattern.

fhembroff:
Created an [issue](https://github.com/dockstore/dockstore/issues/4985) detailing the problems with `DockstoreYaml.readAndValidateDockstoreYaml` and `DockstoreYaml.validateDockstoreYamlProperties`. If these are problems are corrected then this function should catch all incorrect `.dockstore.yml` files.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
fhembroff: I have followed a pattern similar to `PluginClient` and `SearchClient`, however `JCommanderUtility` was causing some issue with my use of the require parameter of `--path <path>` so I did it slightly differently.

dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/YamlClientIT.java
```diff
@@ -0,0 +1,71 @@
...
+    public void missingPathParamter() {
+        Client.main(new String[] { "yaml", "validate" });
+        Assert.assertTrue(systemOutRule.getLog().contains("ERROR: Missing --path <path>"));
+        systemOutRule.clearLog();
+    }
+
+    @Test
+    public void completeRun() {
+        Client.main(new String[] { "yaml", "validate", "--path", "../dockstore-client/src/test/resources/YamlVerifyTestDirectory/correct-directory" });
+        assertTrue(systemOutRule.getLog().contains("src/test/resources/YamlVerifyTestDirectory/correct-directory/.dockstore.yml is a valid yaml file"));
```
denis-yuen, line 58: these should use a string constant imported from your non-test code, that way a developer can quickly tell what test to modify when changing the return message

dockstore-client/pom.xml
```diff
@@ -296,9 +293,21 @@
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-validator</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.media</groupId>
```
denis-yuen, line 301: I believe the version for jersey stuff is inherited and controlled via property in bom, so you should be able to leave this out as in https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/pom.xml#L249-L252

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,108 @@
...
+        JCommander jcPlugin = JCommanderUtility.addCommand(jc, "yaml", commandYaml);
+        CommandYamlValidate commandYamlValidate = new CommandYamlValidate();
+        commandYamlValidate = new CommandYamlValidate();
+        JCommanderUtility.addCommand(jcPlugin, "validate", commandYamlValidate);
+
+        // Not parsing with jc because we know the first command was plugin.  jc's purpose is to display help
+        jcPlugin.parse(argv);
+        try {
+            if (args.isEmpty() || commandYaml.help) {
+                printJCommanderHelpYaml(jc, "dockstore", "yaml");
```
denis-yuen, line 62: "yaml" gets used a lot in this whole PR, should be declared as a constant

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,108 @@
...
+    @Parameters(separators = "=", commandDescription = "Tools used for yaml files")
+    private static class CommandYaml {
+        @Parameter(names = "--help", description = "Prints help for yaml command", help = true)
+        private boolean help = false;
+
+    }
+
+    @Parameters(separators = "=", commandDescription = "List currently activated file provision plugins")
+    private static class CommandYamlValidate {
+        @Parameter(names = "--path", description = "Complete entry path on computer (ex. /home/usr/test)", required = false)
```
denis-yuen, line 101: a follow-up ticket can be created for relative paths (a little more convenient, but not must-have)

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
```diff
@@ -0,0 +1,170 @@
...
+            // Unlikely to ever catch
+            throw new ValidateYamlException("Error reading " + dockstoreYmlPath.toString() + " :\n" + ex.getMessage());
+        }
+
+        // Verify that the Yaml is valid, however does not verify it is valid for dockstore
+        Yaml yaml = new Yaml();
+        Map<String, Object> data = null;
+        try {
+            data = yaml.load(Files.readString(dockstoreYmlPath));
+            out(dockstoreYmlPath + " is a valid yaml file");
```
denis-yuen, line 136: this should be a string constant and referenced here and in your tests

denis-yuen (commented):
some initial thoughts, looking good and will review again when out of draft mode

dockstore-client/pom.xml
fhembroff: If I don't use this dependency I get the following error

```
[INFO] --- maven-dependency-plugin:3.3.0:analyze-only (analyze) @ dockstore-client ---
[WARNING] Used undeclared dependencies found:
[WARNING]    org.glassfish.jersey.media:jersey-media-multipart:jar:2.33:compile
[INFO] Add the following to your pom to correct the missing dependencies: 
[INFO] 
<dependency>
  <groupId>org.glassfish.jersey.media</groupId>
  <artifactId>jersey-media-multipart</artifactId>
  <version>2.33</version>
</dependency>
```

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerify.java
fhembroff: I have changed all the tests so they use string constants defined mostly in YamlVerify

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
fhembroff: I have changed occurenses of `"yaml"`to `YAML` which is defined in `YamlVerify`

dockstore-client/generated/src/main/resources/pom.xml
```diff
@@ -1,379 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
fhembroff: I thought I accidentally checked it in, but it appears this file was already [there](https://github.com/dockstore/dockstore-cli/blob/develop/dockstore-client/generated/src/main/resources/pom.xml), am I fine to delete it in this PR?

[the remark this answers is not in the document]
denis-yuen: Unfortunately, this needs to be left in the codebase undeleted. 
It is automatically generated during the build process but Snyk uses these files to audit our dependencies (snyk doesn't understand Maven properties)

denis-yuen (commented):
looking good, but re-request after the merge conflict is fixed so I can get a look at codacy/sonarcloud, etc.

dockstore-client/generated/src/main/resources/pom.xml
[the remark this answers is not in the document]
fhembroff: I have added it back in. Thanks for the clarification.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,113 @@
...
+import static io.dockstore.client.cli.YamlVerify.YAML;
+
+/**
+ *
+ */
+public final class YamlClient {
+
+    public static final String NO_PATH_FLAG = "ERROR: Missing --path <path>";
+    public static final String VALIDATE_HELP_MESSAGE = "Verifies " + DOCKSTOREYML + " is correct and ensures all required files are present";
+    public static final String PATH_ON_COMPUTER = "Path on computer (ex. /home/usr/Dockstore/test, ~/Dockstore/test, or ../test)";
```
coverbeck, line 41: 1. "Path on computer" -- "on computer" seems redundant
2. What path is it? If it's a .dockstore.yml, then use either use that as an example (/home/usr/Dockstore/.dockstore.yml), or if it's the directory of the .dockstore.yml, say that. Or of it's something else entirely, describe that. But I feel "path" is too general.

coverbeck (commented):
This doesn't build for me locally, and it looks like the CI builds failed too, so I'm going to stop reviewing -- I'd like to be able to navigate in the IDE.

> I have followed a pattern similar to PluginClient and SearchClient, however JCommanderUtility was causing some issue with my use of the require parameter of --path <path> so I did it slightly differently.

That seems odd. I vaguely remember Richard saying something about how we weren't using JCommander correctly, but it seems like you shouldn't have to hand craft help messages that JCommander already does.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
fhembroff: > This doesn't build for me locally, and it looks like the CI builds failed too, so I'm going to stop reviewing -- I'd like to be able to navigate in the IDE.

It appears that the  merging of this [PR](https://github.com/dockstore/dockstore-cli/pull/165) caused some issues. I Have created this [PR](https://github.com/dockstore/dockstore/pull/5004/files) to correct that issue.

fhembroff: I have changed my wording of "Path on computer" to `"Path to .dockstore.yml (ex. /home/usr/Dockstore/test, ~/Dockstore/test, or ../test)"`

fhembroff:
> That seems odd. I vaguely remember Richard saying something about how we weren't using JCommander correctly, but it seems like you shouldn't have to hand craft help messages that JCommander already does.

I fixed this error,  now it uses JCommander in the way it was intended.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private YamlVerifyUtility() {
+    }
+    // Determines if all the files referenced in a list of strings exist
+    private static List<String> filesExist(List<String> paths, String base) {
+        List<String> missingFiles = new ArrayList<>();
+        for (String path : paths) {
+            Path pathToFile = Paths.get(base, path);
+            if (!Files.exists(pathToFile)) {
+                missingFiles.add(pathToFile.toString() + FILE_DOES_NOT_EXIST);
```
coverbeck, line 71: When I run this it detects an error like this:

```
/Users/charlesoverbeck/git/dockstore/dockstore-cli/dockstore-client/src/test/resources/YamlVerifyTestDirectory/2ToolsWithNoName/dockstore.wdl.json does not exist
```

I think it would be more useful to print out relative paths -- the user is working in the confined space of a repository. Just a suggestion, I could see an argument for the absolute path.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -0,0 +1,182 @@
...
+        String dockstoreYmlString = "";
+        try {
+            dockstoreYmlString = Files.readString(dockstoreYmlPath);
+        } catch (IOException ex) {
+            // Unlikely to ever catch
+            throw new ValidateYamlException("Error reading " + dockstoreYmlPath.toString() + " :\n" + ex.getMessage());
+        }
+
+        // Verify that the Yaml is valid, however does not verify it is valid for dockstore
+        Yaml yaml = new Yaml();
```
coverbeck, line 145: It's a bit of stretch for an attack, since it's not running on any of our servers, but I guess in theory somebody could create a malicious dockstore.yml that you fork/pull locally, so I think you should use SafeConstructor like here: https://github.com/dockstore/dockstore/blob/419dd80a42caf46691920c18cee516e371e6c0ec/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java#L208

@david4096 Thoughts on the need to do this in the client?

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -0,0 +1,182 @@
...
+
+
+
+    private YamlVerifyUtility() {
+    }
+    // Determines if all the files referenced in a list of strings exist
+    private static List<String> filesExist(List<String> paths, String base) {
+        List<String> missingFiles = new ArrayList<>();
+        for (String path : paths) {
+            Path pathToFile = Paths.get(base, path);
```
coverbeck, line 69: We don't officially support Windows, so you can ignore this, but I wonder if this will work on a Windows machine. I forget if Java will translate a Linux relative path to a Windows relative path

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -0,0 +1,182 @@
...
+    4. Verify that all files exist
+ */
+
+public final class YamlVerifyUtility {
+
+    public static final String DOCKSTOREYML = ".dockstore.yml";
+    public static final String YAML = "yaml";
+
+    public static final String VALID_DOCKSTORE_YML = " is a valid dockstore yaml file and all required files are present";
+    public static final String INVALID_DOCKSTORE_YML = "Your .dockstore.yml is invalid:\n";
```
coverbeck, line 49: Technically, all the `\n` usages should be `Platform.lineSeparator()`, but we don't officially support Windows, so I think you can ignore this, unless anybody else chimes in favor of.

coverbeck (commented):
I ran this locally and it's looking really good. Mainly some nit-picky comments, some of which you can ignore.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
fhembroff: This a feature I was attempting to have, currently what it is displaying is the directory you entered as the path parameter with the relate path in `.dockstore.yml` appended to it.

The issue is that if I use `~`  it appears that my system automatically resolves that to `/home/fhembroff` and then uses that as the command line parameter.

Here are two examples of the output when being run in `~/Documents/Dockstore/dockstore-cli` on my computer,
```
 java -jar dockstore-client/target/dockstore-client-1.13.0-SNAPSHOT-shaded.jar yaml --path ~/Documents/Dockstore/dockstore-cli/dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present validate

/home/fhembroff/Documents/Dockstore/dockstore-cli/dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present/.dockstore.yml is a valid yaml file
Your file structure has the following errors:
/home/fhembroff/Documents/Dockstore/dockstore-cli/dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present/dockstore.wdl.json does not exist
/home/fhembroff/Documents/Dockstore/dockstore-cli/dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present/Dockstore.cwl does not exist
```

```
java -jar dockstore-client/target/dockstore-client-1.13.0-SNAPSHOT-shaded.jar yaml --path dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present validate

dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present/.dockstore.yml is a valid yaml file
Your file structure has the following errors:
dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present/dockstore.wdl.json does not exist
dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present/Dockstore.cwl does not exist
```

Now given that most users will have the dockstore script in their paths I thought most of them would run this script in the directory of `.dockstore.yml`.

Here is an example of the output when the following is run in `~/Documents/Dockstore/dockstore-cli/dockstore-client/src/test/resources/YamlVerifyTestDirectory/some-files-present ` on my computer.

```
dockstore  yaml --path ./ validate

./.dockstore.yml is a valid yaml file
Your file structure has the following errors:
./dockstore.wdl.json does not exist
./Dockstore.cwl does not exist
```

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
coverbeck: Yes, that's correct. Sorry, I don't know how/why I typed `Platform.lineSeparator()`

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,92 @@
...
+            jcPlugin.parse(argv);
+        } catch (ParameterException ex) {
+            printJCommanderHelp(jc, "dockstore", YAML);
+            out(ex.getMessage());
+            return true;
+        }
+        if (commandYaml.help) {
+            printJCommanderHelp(jc, "dockstore", YAML);
+        } else {
+            switch (jcPlugin.getParsedCommand()) {
```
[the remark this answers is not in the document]
fhembroff: I have corrected this, now if you omit `validate` it will display the standard `yaml` help message, followed by this line,
```
ERROR: No command given
```

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+        return true;
+    }
+
+
+    @Parameters(separators = "=", commandDescription = "Tools used for " + YAML + " files")
+    private static class CommandYaml {
+        @Parameter(names = "--path", description = "Path to " + DOCKSTOREYML + " (ex. /home/usr/Dockstore/test, ~/Dockstore/test, or ../test)", required = true)
+        private static String path = null;
```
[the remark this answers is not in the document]
fhembroff: The reason I used statics here is because `YamlClient.handleCommand` is declared as static and if `CommandYaml` isn't declared as static I run into errors. It appears this was done in [PluginClient](https://github.com/dockstore/dockstore-cli/blob/ac2ac87d1c738376e981b83cbf17d8b9c4b3241e/dockstore-client/src/main/java/io/dockstore/client/cli/PluginClient.java#L131).

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,92 @@
...
+                // fall through
+            }
+        }
+        return true;
+    }
+
+
+    @Parameters(separators = "=", commandDescription = "Tools used for " + YAML + " files")
+    private static class CommandYaml {
+        @Parameter(names = "--path", description = "Path to " + DOCKSTOREYML + " (ex. /home/usr/Dockstore/test, ~/Dockstore/test, or ../test)", required = true)
```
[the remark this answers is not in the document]
fhembroff: Corrected, I went with option (1) it now reads,
```
 --path <path>  Directory that contains .dockstore.yml (ex. /home/usr/Dockstore/test, ~/Dockstore/test, or ../test)
```

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -0,0 +1,182 @@
...
+        // Validates file dockstoreYaml in general
+        DockstoreYaml12 dockstoreYaml12 = null;
+        try {
+            dockstoreYaml12 = DockstoreYamlHelper.readAsDockstoreYaml12(dockstoreYmlString);
+        } catch (Exception ex) {
+            throw new ValidateYamlException(INVALID_DOCKSTORE_YML + dockstoreYmlPath.toString() + CONTAINS_ERRORS + ex.getMessage());
+        }
+
+        // Determines if all referenced files exist
+        allFilesExist(dockstoreYaml12, workflowPath.toString());
```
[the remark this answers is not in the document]
fhembroff: I have added support for `.github` directories. Now `YamlVerifyUtility` will check if it is in a `.github` directory and if it is it will check all of the paths in `.dockstore.yml` relative to its parent directory.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param args
+     */
+    public static boolean handleCommand(List<String> args) {
+        String[] argv = args.toArray(new String[args.size()]);
+        JCommander jc = new JCommander();
+
+        CommandYaml commandYaml = new CommandYaml();
+        JCommander jcPlugin = JCommanderUtility.addCommand(jc, YAML, commandYaml);
+        CommandYamlValidate commandYamlValidate = new CommandYamlValidate();
+        JCommanderUtility.addCommand(jcPlugin, "validate", commandYamlValidate);
```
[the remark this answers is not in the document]
fhembroff: I have created a variable for `validate`.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
fhembroff: Corrected, the line is now
```
       final Yaml safeYaml = new Yaml(new SafeConstructor());
        try {
            safeYaml.load(Files.readString(dockstoreYmlPath));
            out(dockstoreYmlPath + VALID_YAML_ONLY);
        } catch (Exception ex) {
            throw new ValidateYamlException(INVALID_DOCKSTORE_YML + dockstoreYmlPath.toString() + INVALID_YAML + ex.getMessage());
        }
```

fhembroff:
I think the CirceCI tests aren't running at the moment because the OICR system are down for their quarterly update.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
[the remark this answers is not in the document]
coverbeck: It's solvable by having a reference to the Command instance, something like [here](https://github.com/dockstore/dockstore-deploy/blob/f9626a876112559559fc869faabe6985676699f8/deploy-cli/src/main/java/io/dockstore/deploy/cli/Main.java#L93). But it can stay the way it is, although see next paragraph.

I just noticed, the way you run this is:

```
dockstore yaml --path <dir> validate
```

Where as I would expect the pattern to be:

```
dockstore yaml validate --path <dir>
```

That's how other commands work:

```
dockstore workflow launch --entry <entry>

### NOT
dockstore workflow --entry <entry> launch 
```

Which means, ignoring the whole static issue, that the property maybe belongs in `CommandYamlValidate`.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
coverbeck: I think you could determine it out using `Paths.relativize` or something like that, regardless of how the path is specified. But looking at the Linux `find` command, and you're doing what they're doing, so all good -- sorry for the digression.

```
~/temp$ find . -name index.html
./gatling/1.12.0/index.html
./gatling/1.12.2/index.html
./gatling/1.12.0-1.12.2/index.html
~/temp$ find ~/temp -name index.html
/Users/charlesoverbeck/temp/gatling/1.12.0/index.html
/Users/charlesoverbeck/temp/gatling/1.12.2/index.html
/Users/charlesoverbeck/temp/gatling/1.12.0-1.12.2/index.html
```

coverbeck (commented):
Looking good; really good tests, e.g., I saw you added one for the `.github`.

Still one question that I happened to notice about commands & arguments order.

fhembroff:
> It's solvable by having a reference to the Command instance, something like [here](https://github.com/dockstore/dockstore-deploy/blob/f9626a876112559559fc869faabe6985676699f8/deploy-cli/src/main/java/io/dockstore/deploy/cli/Main.java#L93). But it can stay the way it is, although see next paragraph.
> 
> I just noticed, the way you run this is:

I  removed all the static keywords, they were actually causing a significant problem in some of my tests.

I have made it so the way `validate` is called is `dockstore yaml validate --path /path/to/directory`.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Parameters(separators = "=", commandDescription = "Tools used for " + YAML + " files")
+    private class CommandYaml {
+        @Parameter(names = "--help", description = "Prints help for " + YAML + " command", help = true)
+        private boolean help = false;
+    }
+
+    @Parameters(separators = "=", commandDescription = VALIDATE_HELP_MESSAGE)
+    private class CommandYamlValidate {
+        @Parameter(names = "--path", description = "Directory that contains " + DOCKSTOREYML + " (ex. /home/usr/Dockstore/test, ~/Dockstore/test, or ../test)", required = true)
```
coverbeck, line 95: Really nitpicky and ignorable, especially since I've beat this one to death :) 

1. Pick a more informative name than `test`, e.g., `my-workflow-repo`, `workflow-repo`, `biotool`, etc. (you're not running this on a random .dockstore.yml somewhere on your system, but in the context of a .dockstore.yml that's part of tree that is or is going to be in GitHub.
2. Give a .github example.

coverbeck (approved):
This is great!

There could be a follow-up ticket(s) to document this and/or have a blog post about it.

coverbeck:
I approved, but you may want to take a look at the Codacy and SonarCloud warnings.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlClient.java
fhembroff: Changed it so the help message is now,
```
Required parameters:
  --path <path>  Directory that contains .dockstore.yml (ex. /home/usr/workflow-repo, ~/Dockstore/.github, or ../biotool)
```

david4096 (approved):
Great test coverage and a huge improvement over the current experience of authoring the .dockstore.yml. How will these errors be surfaced in the UI? In the github apps error logs?

fhembroff:
> How will these errors be surfaced in the UI? In the github apps error logs?

Since these changes are only in CLI and not being used in the UI no error messages will be displayed in the UI.

fhembroff:
Realize I never mentioned the github issue on this PR. Mentioning it now https://github.com/dockstore/dockstore/issues/4838

PR #165, opened 2022-06-22
Use BOM versions of jackson core etc.

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-4477

The cron command is 
```
./mvnw clean test -Punit-tests -Ddockstore-core.version=1.13.0-20220621.170042-53 -B -ntp -U
```
The -U switch [forces a check for missing releases and updated snapshots on remote repositories](https://maven.apache.org/ref/3.6.3/maven-embedder/cli.html) and is what caught the version collision on GitHub actions.

Perhaps we should use the -U switch when building locally, e.g. 
```
./mvnw -U clean install
```


pom.xml
```diff
@@ -137,18 +137,6 @@
...
-
-            <!--
-                  override the BOM for jackson-core version
-                  until the BOM version is upgraded
-                  several dependencies in the CLI need the newer version
-             -->
-            <dependency>
-                <groupId>com.fasterxml.jackson.core</groupId>
-                <artifactId>jackson-core</artifactId>
-                <version>2.13.1</version>
```
denis-yuen, line 149: Rather than downgrade, could also update to newer bom 
https://github.com/dockstore/dockstore/blob/develop/bom-internal/pom.xml#L80

I think we upgraded to satisfy some snyk checks

denis-yuen (commented):
> Perhaps we should use the -U switch when building locally, e.g.

Nice find!

Not sure about locally since it would presumably be slower since it would need to contact the remote repository every time, but we might want to consider getting `-U` into some of the other github checks like https://github.com/dockstore/dockstore-cli/blob/develop/.github/workflows/mvnw.yml#L32-L33 or https://github.com/dockstore/dockstore/blob/develop/.github/workflows/mvnw.yml#L32-L33

coverbeck (commented):
Was here, waiting for response to Denis' suggestion.

svonworl (approved):

pom.xml
wshands: OK, local test with webservice develop branch build 1.13.0-SNAPSHOT which pulls in the upgraded BOM works. 
`<dockstore-core.version>1.13.0-SNAPSHOT</dockstore-core.version>`

CLI generated pom.xml on my local machine
```
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-core</artifactId>
      <version>2.13.2</version>
      <scope>compile</scope>
```

However the latest webservice build on Artifactory was uploaded 2/09/2022 which is before the  jackson-core version was upgraded to 2.13.2.

Screen Shot 2022-06-22 at 2 31 37 PM

Screen Shot 2022-06-22 at 2 49 40 PM

and [current CLI dockstore core is 1.13.0-alpha.6](https://github.com/dockstore/dockstore-cli/blob/9a53438aec6e21e6f322bbe407e85ef78341049f/pom.xml#L41)
Screen Shot 2022-06-22 at 2 53 00 PM

I'm not sure what the best solution is, maybe bump up jackson version in CLI pom?

denis-yuen: I would suggest making a new build, we're not really running out of alphas :wink:

wshands: I will make a new webservice release and change the CLI pom

wshands: Created a new webservice build which has the BOM with upgraded jackson-core version and pointed the CLI to it.

denis-yuen (approved):

fhembroff (approved):

coverbeck (approved):
We chatted about this today; there was a bit more to this PR than I realized, so it's probably just as well to merge it, but there should be a followup ticket to restore the ignored tests.

PR #171, opened 2022-07-15
Feature/seab 4525/bitbucket split

fhembroff:
## UPDATE
Whilst the below description remains valid, I have introduced some additional changes.

What I have done is delete the BitBucket token from the database whenever `resetDBBetweenTests()` is run (except in the case of the BitBucket Tests). This is required as some tests still contact BitBucket even though they do not require BitBucket.
 
To see just how many tests used BitBucket when they ran I deleted the BitBucket token from `circle_ci_test_data.zip.enc` and ran it in CircleCI. Results found [here](https://app.circleci.com/pipelines/github/dockstore/dockstore-cli/1202/workflows/adea4032-1366-4446-a750-087b861bd1a6/jobs/4795).

Then after implementing the changes in this PR (but still without the BitBucket token) I ran the CircleCI job again and got this [result](https://app.circleci.com/pipelines/github/dockstore/dockstore-cli/1203/workflows/b90b2c39-faf8-4d4c-a6cd-79ac4864b614). Which shows that all the tests that failed before (except those in BitBucket test files) passed.

This change will ensure that when BitBucket stops responding, only tests in the BitBucket jobs will fail.

I will create a similar PR for https://github.com/dockstore/dockstore shortly.

## Description
This PR splits out all tests that use BitBucket, creates a new profile for them and creates a new job for them in CircleCI.

### Issue
[SEAB-4525](https://ucsc-cgl.atlassian.net/browse/SEAB-4525)

## Specific Changes
The following tests were moved because they use the username `dockstore_testuser2` which uses the URL [bitbucket.org](bitbucket.org) as it's `tokensource` in `dockstore-webservice/src/main/resources/migrations.test.confidential2.xml`.
 - `GeneralWorkflowIT.testRefreshAndConvertWithImportsWDL` 
 - `GeneralWorkflowIT.testManualPublishBitbucket`
 - `GeneralWorkflowIT.testBitbucketDirtyBit`

The following tests were moved because they use the name `quayandbitbucket` which uses the giturl `git@bitbucket.org:DockstoreTestUser/dockstore-whalesay.git` in `dockstore-webservice/src/main/resources/migrations.test.confidential1.xml`.
 - `BasicIT.testQuayAndBitbucketPublishAndUnpublishAentry`
 - `BasicIT.testRefreshingUserMetadata`
 - `BasicIT.testRefreshCorrectTool`
 - `BasicIT.testQuayBitbucketManuallyRegisterDuplicate`

The following tests were moved because they use the name `quayandbitbucketalternate` which uses the giturl `git@bitbucket.org:DockstoreTestUser/quayandbitbucketalternate.git` in `dockstore-webservice/src/main/resources/migrations.test.confidential1.xml`.
 - `BasicIT.testQuayBitbucketPublishAlternateStructure`
 - `BasicIT.testQuayBitbucketManualPublishAndUnpublishAlternateStructure`
 - `BasicIT.testDockerhubBitbucketAlternateStructure`
 - `BasicIT.testDockerhubBitbucketWrongStructure`

The following tests were moved they have `git@bitbucket.org` in their giturl specified in `Client.main(...)`
 - `BasicIT.testDockerhubBitbucketManualRegistrationDuplicates`
 - `BasicIT.testDockerhubBitbucketManualRegistration`
## Checks
 - [ ] Check that you pass the basic style checks and unit tests by running `mvn clean install`


coverbeck (approved):

svonworl (approved):

denis-yuen (approved):
on passing builds

fhembroff:
Making some large changes. I expect that a few tests may fail for this PR in the next hour or two.

fhembroff:
Made some significant changes. Updated PR description.

coverbeck:
> This is required as some tests still contact BitBucket even though they do not require BitBucket.

Do you know where exactly this is happening?

fhembroff:
> > This is required as some tests still contact BitBucket even though they do not require BitBucket.
> 
> Do you know where exactly this is happening?

I believe it happens when `containersApi.refresh();` is called.

coverbeck (approved):
Nice find! The other benefit to this fix is that will also reduce number of Bitbucket API calls, making it less likely for the limit to be reached, although I don't know if it's significant enough.

One suggestion though, but see what others think first, but I think it would be cleaner if tests had to "opt-in" to getting the Bitbucket token, rather than all the non-Bitbucket tests having to opt out.

It's probably more work than it's worth to split up the encoded DB, but you could add 
`CommonTestUtilities.cleanStatePrivate1(SUPPORT, needBucketToken)` method or something like that, and all the other `cleanState...` methods that don't have the param or have it set to false, delete the token. But this is minor.

kathy-t (approved):
I like Charles' suggestion of having BitBucket tests opt-in to have the BitBucket token rather than having all the non-Bitbucket tests opt out. If there was a new non-Bitbucket `IT` class, the developer may not know/forget to opt out of the BitBucket token

fhembroff:
> I like Charles' suggestion of having BitBucket tests opt-in to have the BitBucket token rather than having all the non-Bitbucket tests opt out. If there was a new non-Bitbucket `IT` class, the developer may not know/forget to opt out of the BitBucket token

Okay, I will implement this suggestion. The build will temporarily break for this PR as I will be verifying that the change still works as intended. I'll make another reply when it is fixed.

denis-yuen (commented):
Looking good, good feedback above

svonworl (approved):

fhembroff:
I have implemented the PR feedback and verified that the changes are working as expected.

denis-yuen (approved):

PR #174, opened 2022-07-21
Removed Liquibase Changeset Message from Output in CircleCI

fhembroff:
**Description**
An update in liquibase has caused every changeset to be logged to stdout. Including messages such as,
```
Running Changeset: migrations.1.12.0.xml::topicPhaseTwoChanges::gluu (generated)
Running Changeset: migrations.1.12.0.xml::clearCwlToolTableJson::svonworl
Running Changeset: migrations.1.12.0.xml::checksumsForNullContent::coverbeck
Running Changeset: migrations.1.12.0.xml::addAppToolEvents::dyuen (generated)
Running Changeset: migrations.1.12.0.xml::orcidAuthorInfo::ktran (generated)
Running Changeset: migrations.1.13.0.xml::migrateBioToBlogLink::ktran
Running Changeset: migrations.1.13.0.xml::convertUserProfileEmptyStringToNull::ktran
Running Changeset: migrations.1.13.0.xml::add_constraints_to_user_profile::ajandu
```
which is making it difficult to determine the cause of test failures in CirlceCI. This issue has been bought to the attention of liquibase and it looks like they will fix it,

https://github.com/liquibase/liquibase/issues/2396

In the meantime I have added a `grep` to each command that produces liquibase changeset messages so that they aren't displayed in CircleCI.

This PR should be revisited when https://github.com/liquibase/liquibase/issues/2396 is resolved.

**Issue**
[SEAB-4602](https://ucsc-cgl.atlassian.net/browse/SEAB-4602)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,7 +339,7 @@ commands:
     steps:
       - run:
           name: run tests
-          command: mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate $MAVEN_GOAL -P$TESTING_PROFILE,coverage -ntp
+          command: mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate $MAVEN_GOAL -P$TESTING_PROFILE,coverage -ntp | grep  -v "^Running Changeset:"
 
   save_test_results:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -340,7 +340,7 @@ commands:
       - run:
           name: run tests
           command: mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate $MAVEN_GOAL -P$TESTING_PROFILE,coverage -ntp | grep  -v "^Running Changeset:"
-
+          # The piping grep command is a temporary fix to this issue https://github.com/liquibase/liquibase/issues/2396
   save_test_results:
     steps:
       - run:


fhembroff:
This PR is near identical to https://github.com/dockstore/dockstore/pull/5033

.circleci/config.yml
```diff
@@ -339,8 +339,8 @@ commands:
     steps:
       - run:
           name: run tests
-          command: mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate $MAVEN_GOAL -P$TESTING_PROFILE,coverage -ntp
-
+          command: mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate $MAVEN_GOAL -P$TESTING_PROFILE,coverage -ntp | grep  -v "^Running Changeset:"
```
coverbeck, line 342: If there's an error applying a change set (rare, I know), will this suppress the error message as well?

coverbeck (approved):
Thought of an edge case, but I don't know that it's that crucial.

.circleci/config.yml
fhembroff: As far as I can tell whenever there is an error when applying a change set one of two things happens,

1. The outputted line does not start with `Running Changeset` and instead starts with `ERROR` (meaning it will not be removed), such was the case in this error example,
 ```
ERROR [2022-07-21 23:50:20,831] liquibase.parser.core.xml.XMLChangeLogSAXParser$1: cvc-complex-type.2.4.a: Invalid content was found starting with element ...
 ```
2. At least two lines are outputted, the first one starts with `Running Changeset` (which will be removed), but the following lines explain the error and provide context (which aren't cropped out), such was the case in this error example,

```
Running Changeset: home/fhembroff/Documents/Dockstore3/dockstore-cli/dockstore-cli-integration-testing/../dockstore-webservice/src/main/resources/migrations.test.confidential2.xml::1511391261451-3::dyuen (generated)
liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set home/fhembroff/Documents/Dockstore3/dockstore-cli/dockstore-cli-integration-testing/../dockstore-webservice/src/main/resources/migrations.test.confidential2.xml::1511391261451-3::dyuen (generated):
     Reason: liquibase.exception.DatabaseException: ERROR: column "blue" does not exist
```

Either way information about the error and it's context is displayed. So I don't think this will affect our ability to debug changeset error messages.

\* NOTE: In both code example there was significantly more output.

kathy-t (approved):

denis-yuen (approved):

PR #209, opened 2022-11-30
SEAB-5021:  Confirm support for CWL Version 1.2

svonworl:
**Description**
This PR adds a couple of tests that help to confirm CLI compatibility with CWL version 1.2.

Per the spec, CWL changed as follows between version 1.1 and 1.2:
https://www.commonwl.org/v1.2/Workflow.html#Changelog
Not much is different.

Knowing how we parse and a bit about how the CLI is structured, my theory was that if our CLI supports CWL v1.1, and the CLI could process a v1.2 workflow that contained either a "when" clause or an "Operation", that it probably can successfully process a very large proportion of v1.2 workflows.  So, I added a test corresponding to each of those two cases.

As it turns out, the CLI processes a workflow with a step containing an "Operation" with no problems, but when cwltool is invoked to run it, it fails because an "Operation" is defined as a placeholder with undefined semantics which cannot be executed:
https://www.commonwl.org/v1.2/Workflow.html#Operation
In other words, this is expected behavior.

**Review Instructions**
Confirm the presence of the new tests, and that they pass.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5021

### DIFF
--- a/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
+++ b/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
@@ -125,7 +125,7 @@ public class LaunchTestIT {
     }
 
     @Test
-    public void cwlHandlesVersion12() {
+    public void cwlSupportsVersion12() {
         // 1st-workflow-12.cwl is a version of 1st-workflow.cwl to which key new CWL version 1.2 features have been added.
         File cwlFile = new File(ResourceHelpers.resourceFilePath("1st-workflow-12.cwl"));
         File cwlJSON = new File(ResourceHelpers.resourceFilePath("1st-workflow-job.json"));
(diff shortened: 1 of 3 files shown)


coverbeck (approved):

svonworl:
> > An implementation (or other tooling) may provide a means of binding an Operation to a concrete process (such as Workflow, CommandLineTool, or ExpressionTool) with a compatible signature.
> 
> Interesting find, did you find any examples? Is this like CWL having its own concept of inheritance? If so, would it be worthwhile testing a CWL workflow with a concrete process that implements(?) an operation?

I _think_ an Operation is more like a placeholder:
https://www.commonwl.org/user_guide/topics/operations.html

So, I'd say no, we probably don't need another test because a fully-cooked, reproducible workflow shouldn't contain Operations.  If/when we tighten up CWL validation in the webservice, we might want to consider marking a workflow as invalid if it contains an Operation step, perhaps.

denis-yuen (approved):

kathy-t (approved):

PR #214, opened 2023-01-10
Feature/junit5

denis-yuen:
**Description**
CLI version of the chain of PRs that ended with https://github.com/dockstore/dockstore/pull/5301

There are still a couple tests that use PowerMock (seems dead, doesn't support Java 17 or JUnit 5, generally bad practice) that are more involved and will require a spin-offf PR before junit4 can be banned from the CLI as well. Also some tests for cwl launching.

One obnoxious test took forever, a reminder to check your assumptions. It was deterministically running differently on CircleCI as opposed to locally, but changing whether it failed and passed on unrelated changes. Looks like the plugin directory was being downloaded and/or cached, but only when some random tests (probably in completely other classes) had run before the test in question, whether due to distribution by the parallel tests script or some oddity due to CircleCI caching. 

This is a long way of saying, check your assumptions about tests and how they're isolated from one another. Is the filesystem reset, have stdout and stderr been reset, etc. Oy. 

**Review Instructions**
Number of tests should be the same (or more) before/after this PR
Hmmm, right now at time of comment, it looks like we go from 391 (insights) to 402 (manual count, may need wait for 24 hours for insights to update)

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5133

**Security**
None known


svonworl (approved):
Looks good!

Out of curiosity, a lot of methods/classes changed from `public` to package access, what was the reason?

denis-yuen:
> Looks good!
> 
> Out of curiosity, a lot of methods/classes changed from `public` to package access, what was the reason?

Junit 5 added the ability to use package access. https://github.com/junit-team/junit5/issues/679 
Sonar Cloud/Codacy suggested it plus since we never/rarely(?) export tests and use them from a different package, it seemed appropriate.

kathy-t (approved):

hyunnaye (approved):

coverbeck (approved):

PR #215, opened 2023-01-12
Added a  ToilCompatibleTest class to CLI

fhembroff:
**Description**
The CLI requires the use of the `ToilCompatibleTest` class. However, `ToilCompatibleTest` class was recently removed from `dockstore/dockstore` in this [PR](https://github.com/dockstore/dockstore/pull/5297). Meaning that the CLI cannot run off the current `dockstore/dockstore` repo. This PR adds the `ToilCompatibleTest` to the CLI.

**Review Instructions**
To review this ticket, you should

- [ ] Verify that the CLI builds correctly with the current `dockstore/dockstore` repo

**Security**

\

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -725,6 +725,11 @@ public class Client {
                         argsArray = args.toArray(argsArray);
                         handled = DepCommand.handleDepCommand(argsArray);
                     } else if (YAML.equals(mode)) {
+                        if (!DEBUG.get() && !INFO.get()) {
+                            // Switching log off if DEBUG or INFO is not set
+                            // yamlClient will generate unnecessary error logs
+                            root.setLevel(Level.OFF);
+                        }
                         yamlClient = new YamlClient();
                         handled = yamlClient.handleCommand(args);
                     }
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -725,11 +725,6 @@ public class Client {
                         argsArray = args.toArray(argsArray);
                         handled = DepCommand.handleDepCommand(argsArray);
                     } else if (YAML.equals(mode)) {
-                        if (!DEBUG.get() && !INFO.get()) {
-                            // Switching log off if DEBUG or INFO is not set
-                            // yamlClient will generate unnecessary error logs
-                            root.setLevel(Level.OFF);
-                        }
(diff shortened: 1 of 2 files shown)


fhembroff:
I do not have any reason to believe that this PR would cause bitbucket tests to fail. So, I will leave it as is.

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -725,6 +725,11 @@ private void run(String[] argv) {
                         argsArray = args.toArray(argsArray);
                         handled = DepCommand.handleDepCommand(argsArray);
                     } else if (YAML.equals(mode)) {
+                        if (!DEBUG.get() && !INFO.get()) {
+                            // Switching log off if DEBUG or INFO is not set
+                            // yamlClient will generate unnecessary error logs
+                            root.setLevel(Level.OFF);
```
denis-yuen, line 731: Kinda wonder if you can set the log level for a particular package or lower in the tree. Probably not too important since this only affects this specific command

denis-yuen (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
coverbeck: I didn't know this until looking at this PR, but there's already code above starting around line 677 that invokes `root.setLevel()` based on the command line switches, and it feels weird to me to set it to `Level.OFF` only for this particular path.

I'd suggest one of these:

1. Doing `root.setLevel(Level.OFF)` up above instead.
2. Don't use LOG.error for errors, use LOG.info
3. Do a pattern something like this:

```
if (LOG.isDebugEnabled()) { 
   LOG.error("the error message", theException);
}
```

I dislike 2.

I lean towards 3 -- it seems a little counter-intuitive, but the nice thing about it is that it won't log by default, but if it does log, it's logging at the ERROR level, so it will stand out for developers. 1 seems a little severe, it's also possible.

Also, see dockstore/dockstore#5223, which relates to this one -- ideally the solution would work for both.

coverbeck (commented):
Left a comment.

denis-yuen (approved):
:laughing: 
not much here now

coverbeck (commented):
While this change is good, the PR title nor linked issue seem applicable. Could you please either fold into your other PR, or update the title/description of this PR?

fhembroff:
> While this change is good, the PR title nor linked issue seem applicable. Could you please either fold into your other PR, or update the title/description of this PR?

I have updated the title and description.

kathy-t (approved):

coverbeck (approved):

PR #218, opened 2023-01-18
DOCK-2219: removed upgrade to unstable version message

hyunnaye:
**Description**
This PR removes the `upgrade-unstable` message in both `dockstore --upgrade` and `dockstore --version`

**Review Instructions**
Review that the message is removed accordingly.

**Issue**
https://github.com/dockstore/dockstore/issues/5054

**Security**


coverbeck (approved):
It would be nice if there were a test for this, but my guess it's going to be pretty hard and not worth the effort, as the test would need to download the JAR and the current version is not a constant.

PR #224, opened 2023-02-14
SEAB-5247: Add MuteForSuccessfulTests extension to CLI

svonworl:
**Description**
This PR adds our new `MuteForSuccessfulTests` Junit 5 extension to the CLI tests.

Important changes include:
1. changing `dockstore-core.version` to `1.14.0-alpha.7` (to reference the new extension).
2. tweaks to the extension annotations and their order (because they were missing or redundant with annotations in parent classes, or to change the order so the "test status" is printed last).
3. addition of the `confidential4` paths to `.gitignore`, even though they're not included in the bundle, because I'd added them for another change that got reverted, but there's no harm in leaving them.

**Review Instructions**
Find a failed CircleCI run that incorporates these changes and confirm that the stdout/stderr output is included with, and only with, the failures. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5247

**Security**


dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
```diff
@@ -50,8 +51,9 @@
  * Base integration test class
  * A default configuration that cleans the database between tests
  */
-@ExtendWith(BaseIT.TestStatus.class)
 @ExtendWith(SystemStubsExtension.class)
+@ExtendWith(MuteForSuccessfulTests.class)
+@ExtendWith(BaseIT.TestStatus.class)
```
denis-yuen, line 56: `BaseIT` redundant here, but probably not worth fixing unless there are other issues

denis-yuen (approved):
which there are not, thanks!

coverbeck (approved):

fhembroff (approved):

PR #233, opened 2023-03-27
Fix Yaml Validation for when testParameterFiles is empty or non-existent

fhembroff:
**Description**
This PR makes it such that you can have a `.dockstore.yml` file that has either an empty or non-existent `testParameterFiles` field. This is valid and accepted by dockstore.org, but it wasn't accepted by the dockstore yaml validator (until now).

Whilst completing this PR I also realized `DockstoreYamlHelper.validateDockstoreYamlProperties()` did not throw an error when

- a `.dockstore.yml` file contained no `primaryDescriptorPath`
- a service contained no files

The former was already fixed, but for the ladder was not, hence why I added this [code](https://github.com/dockstore/dockstore-cli/blob/2fce8d66433f487dd4864257eaa48293c5cf3de2/dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java#L122-L126). I have created this [ticket](https://github.com/dockstore/dockstore/issues/5443) to improve `DockstoreYamlHelper.validateDockstoreYamlProperties()` and left instructions to remove the fixes in the CLI for the above two errors when the issue is resolved (as they should be caught by `DockstoreYamlHelper.validateDockstoreYamlProperties()`).

**Review Instructions**
Ensure that a `.dockstore.yml` file with either an empty or non-existent `testParameterFiles` field (but with everything else correct) passes the dockstore yaml validator with no errors. As it is valid to have a `.dockstore.yml` file with either an empty or non-existent `testParameterFiles` field.

**Issue**
https://github.com/dockstore/dockstore/issues/5401
[DOCK-2338](https://ucsc-cgl.atlassian.net/browse/DOCK-2338)

**Security**

[DOCK-2338]: https://ucsc-cgl.atlassian.net/browse/DOCK-2338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


denis-yuen (commented):
A couple minor questions/nitpicks, approach in general seems ok

dockstore-client/src/test/resources/YamlVerifyTestDirectory/service-empty-files-parameter/.dockstore.yml
```diff
@@ -0,0 +1,21 @@
...
+  scripts:
+    start: start.sh
+  files:
+  environment:
+    TEST-1:
+      default: testing
+      description: testing testing testing
+  data:
+    test-2:
+      targetDirectory: ~/test
```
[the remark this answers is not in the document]
fhembroff: This would've just been me filling out the template. I just copied a `.dockstore.yml` file I had already created for this file.

However, after reading through the [template](https://docs.dockstore.org/en/stable/assets/templates/services/services.html#full-template-with-explanation-of-all-available-fields), it appears this is not valid, as it requires that the `targetDirectory` is a relative path.

I have corrected this for all `.dockstore.yml` test files that I created.

I have also added this [comment](https://github.com/dockstore/dockstore/issues/5443#issuecomment-1487089867) to this [ticket](https://github.com/dockstore/dockstore/issues/5443) to either determine correct `DockstoreYamlHelper.validateDockstoreYamlProperties()` such that it catches this in the future, or to update the [template](https://docs.dockstore.org/en/stable/assets/templates/services/services.html#full-template-with-explanation-of-all-available-fields) to indicate that absolute paths are acceptable.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -76,40 +77,56 @@ private static List<String> filesExist(List<String> paths, String base) {
...
-                List<String> filePaths = tool.getTestParameterFiles();
-                filePaths.add(tool.getPrimaryDescriptorPath());
-                missingFiles.addAll(filesExist(filePaths, basePath));
+                List<String> testParameterFiles = tool.getTestParameterFiles();
+
+                if (testParameterFiles != null) {
+                    filePathsToCheck.addAll(testParameterFiles);
+                }
+
+                filePathsToCheck.add(tool.getPrimaryDescriptorPath());
```
coverbeck, line 114: It was like this, but this whole code block duplicates code block starting at line 95. You could optionally refactor a couple of ways to avoid the duplication.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -76,40 +77,56 @@ private static List<String> filesExist(List<String> paths, String base) {
...
-            missingFiles.addAll(filesExist(service.getFiles(), basePath));
+            List<String> files = service.getFiles();
+            if (files != null) {
+                filePathsToCheck.addAll(files);
+            } else {
+                missingFiles.add(SERVICE_DOES_NOT_HAVE_FILES);
+            }
         }
+
+        missingFiles.addAll(filesExist(filePathsToCheck, basePath));
```
coverbeck, line 129: Comment is for line 122/139, and it was like this, but `INVALID_FILE_STRUCTURE`, which translates to `Your file structure has the following errors:` -- I would say something like `Your .dockstore.yml has the following errors:` -- INVALID_FILE_STRUCTURE is slightly misleading variable name too (the structure of the file is fine in this case, it's just values that aren't good).

coverbeck (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
fhembroff: Refactored it a bit.

Thanks for the feedback.

PR #238, opened 2023-05-08
dropwizard 4 update

denis-yuen:
**Description**
Skips dropwizard 3 and goes directly to dropwizard 4
CLI counterpart to https://github.com/dockstore/dockstore/pull/5483 and likely with many of the same issues

May or may not create issue to update file provisioning plugins. Having the old jaxb around seems enough to get them to work, don't know if that will eventually cause issues (although they are simpler than swagger/openapi client libraries)

**Review Instructions**
CLI should work in general (spot check a few commands like validating things, talking to dockstore.org, etc.)

**Issue**
https://github.com/dockstore/dockstore/issues/5492  (dropwizard 3 update)
https://ucsc-cgl.atlassian.net/browse/SEAB-5454 json update
https://ucsc-cgl.atlassian.net/browse/SEAB-5468 (dropwizard 4 update)

**Security**


coverbeck (approved):

pom.xml
```diff
@@ -532,12 +532,22 @@
                                         <exclude>javax.servlet:javax.servlet-api</exclude>
                                         <!-- Replaced with jakarta.validation:jakarta.validation-api -->
                                         <exclude>javax.validation:validation-api</exclude>
-                                        <!-- Replaced with jakarta.xml.bind:jakarta.xml.bind-api -->
-                                        <exclude>javax.xml.bind:jaxb-api</exclude>
+                                        <!-- Replaced with jakarta.xml.bind:jakarta.xml.bind-api, but need to keep around for old plugins -->
+                                        <!-- <exclude>javax.xml.bind:jaxb-api</exclude> -->
```
kathy-t, line 536: Is this commented out exclusion needed?

kathy-t (approved):

pom.xml
denis-yuen: Kinda. 
This is a result of  

> May or may not create issue to update file provisioning plugins. Having the old jaxb around seems enough to get them to work, don't know if that will eventually cause issues (although they are simpler than swagger/openapi client libraries)

Ideally, we'd update the plugins and then add this ban. Allowing this library and the new jarkarta ones at the same time seems to work right now, but may lead to problems in the future. 
In other words, this could be considered aspirational.

david4096 (approved):
Here's to hoping dropwizard doesn't have to change out their backing library again soon.

denis-yuen:
> Here's to hoping dropwizard doesn't have to change out their backing library again soon.

Java rather, but agreed

PR #255, opened 2023-08-23
DOCK-2262: invalidates relative path yml files

hyunnaye:
**Description**
This PR invalidates `./dockstore.yml` files that provide relative paths instead of absolute.
**Review Instructions**
Verify that the new test pass and yml files with relative paths are not valid anymore using the webservice validator constraints.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2262

**Security**


denis-yuen (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -80,6 +81,15 @@ private static List<String> filesExist(List<String> paths, String base) {
         return missingFiles;
     }
 
+    // Determines if all the file paths are absolute
+    private static boolean checkAbsoluteFilePaths(List<String> paths) {
+        for (String path : paths) {
+            if (path.length() != 0 && !path.startsWith("/")) {
```
coverbeck, line 87: Nit/optional, instead of `path.length() != 0`, you could have `!path.isEmpty()`. The implementation `isEmpty()` is your code, but `isEmpty()` arguably better conveys what you're trying to do.

dockstore-client/src/test/resources/YamlVerifyTestDirectory/relative-primary-descriptor-path/Dockstore2.wdl
coverbeck, line 1: Do you need this file? You reference it, but with a non-absolute path...

coverbeck (approved):

dockstore-client/src/test/java/io/dockstore/client/cli/YamlValidatorTest.java
```diff
@@ -83,6 +83,18 @@ void emptyDockstoreYml() {
         }
     }
 
+    //Determins if .dockstore.yml provides relative paths
+    @Test
+    void relativePathYml() {
```
kathy-t, line 88: Could also add a test for test parameter files with relative paths

kathy-t (approved):

dockstore-client/src/test/resources/YamlVerifyTestDirectory/relative-primary-descriptor-path/Dockstore2.wdl
hyunnaye: I was testing that the non-absolute path reference fails. But I see the confusion so I'll add one more successful test

svonworl (commented):
If we also want to enforce absolute paths in the webservice, I'd recommend that we implement the check as a `Validator` in dockstore-common.  For example, you could have an `AbsolutePath` validator, and annotate the classes that represent the parsed yaml, for example `YamlWorkflow`:

```
    @NotNull
    @AbsolutePath
    public String getPrimaryDescriptorPath() {
        return primaryDescriptorPath;
    }
```

This way, paths in both the webservice and the yaml validator are checked by the same code.

dockstore-client/src/test/resources/YamlVerifyTestDirectory/relative-primary-descriptor-path/Dockstore2.wdl
kathy-t: Hmm, I think Charles meant that you might be able to delete this file because the validation would fail before this file is ever read, but looking at the code again, it seems like we check that the file exists before validating that all paths are absolute. If we validated absolute paths first, then you wouldn't need to have this file for your test.

dockstore-client/src/main/java/io/dockstore/client/cli/YamlVerifyUtility.java
```diff
@@ -131,6 +136,9 @@ private static void allFilesExist(DockstoreYaml12 dockstoreYaml12, String basePa
             String errorMsg = errorMsgBuild.toString();
             throw new ValidateYamlException(MISSING_FILE_ERROR + errorMsg);
         }
+        if (!checkAbsoluteFilePaths(filePathsToCheck)) {
+            throw new ValidateYamlException(RELATIVE_PATH_ERROR);
+        }
```
kathy-t, line 141: Realized this after - it might make more sense to move this absolute file check to line 121/126 before `missingFiles.addAll(filesExist(filePathsToCheck, basePath));`. We can avoid checking if files exist if the files have relative paths

svonworl (approved):

hyunnaye:
Rerequesting reviews for everybody as I removed all previous code and let the cli use the new webservice constraints from https://github.com/dockstore/dockstore/pull/5641. This PR is now just to add some tests to verify that the absolute paths are not valid.

denis-yuen (approved):
Thanks for persevering!

dockstore-client/src/test/java/io/dockstore/client/cli/YamlValidatorTest.java
```diff
@@ -83,6 +83,32 @@ void emptyDockstoreYml() {
...
+            assertTrue(ex.getMessage().contains("the path must be an absolute path to be valid"));
+        }
+
+    }
+
+    //Verifies .dockstore.yml is invalid when provided relative test descriptor paths
+    @Test
+    void relativeTestDescriptorPathYml() {
+        final String relativePathDockstoreYmlDirectory = "src/test/resources/YamlVerifyTestDirectory/relative-test-files-path";
+        try {
```
coverbeck, line 103: Lines 103-108 are a duplicate of 90-95; you could create a common method.

coverbeck (approved):

svonworl (approved):

dockstore-client/src/test/resources/YamlVerifyTestDirectory/relative-test-files-path/Dockstore2.wdl
kathy-t, line 1: Do you need to have this empty file? I know you reference it in the .dockstore.yml but my theory is that you can remove this file and the .dockstore.yml absolute path validation would still work as expected because the validation happens before the file is read

dockstore-client/src/test/resources/YamlVerifyTestDirectory/relative-test-files-path/dockstore.wdl.json
kathy-t, line 1: Ditto about removing this file

dockstore-client/src/test/resources/YamlVerifyTestDirectory/relative-primary-descriptor-path/Dockstore2.wdl
kathy-t, line 1: Ditto about removing this file

kathy-t (approved):

PR #268, opened 2023-12-01
Move nightly build to CircleCI

coverbeck:
**Description**
Moves nightly build against dockstore/dockstore develop snapshot to CircleCI from GitHub actions.

The bug noted that the nightly build against develop was only catching compilation errors. This turned out to be because the GitHub action was only running the unit tests. It was not running the integration tests.

To add the integration tests, it was decided to move the build to CircleCI. This avoids duplication, e.g., decryption for confidential tests, invoking the different Maven profiles, etc.

The trigger to fire this nightly is [here](https://app.circleci.com/settings/project/github/dockstore/dockstore-cli/triggers?return-to=https%3A%2F%2Fapp.circleci.com%2Fprojects%2Fproject-dashboard%2Fgithub%2Fdockstore%2F) -- will finalize it once this is merged.

**Review Instructions**
1. Verify that a build is run nightly against the lastest snapshot of dockstore/dockstore
2. Verify that "regular" builds are building against the version in the pom.xml

To do so, you need to look at the build logs, and look for dockstore-common. 

The build against develop webservice message will look like this, note the `SNAPSHOT`:

```
Downloading from artifacts.oicr.on.ca.ca-snapshots: https://artifacts.oicr.on.ca/artifactory/collab-snapshot/io/dockstore/dockstore-common/1.15.0-SNAPSHOT/dockstore-common-1.15.0-20231201.202101-134.jar
```

The build against the web service version defined in the pom will look like this:

```
Downloading from artifacts.oicr.on.ca.ca-snapshots: https://artifacts.oicr.on.ca/artifactory/collab-snapshot/io/dockstore/dockstore-common/1.15.0-alpha.11/dockstore-common-1.15.0-alpha.11.pom
```

**Issue**
SEAB-5853

**Security**


.circleci/config.yml
```diff
@@ -30,60 +34,99 @@ common_filters: &common_filters
...
-            - confidential-workflow-tests
-            - confidential-tool-tests
-            - wes-toil-test
-          context:
-            - sonarcloud
+    when:
+      not: << pipeline.parameters.run_against_develop_core >>
+    <<: *common_jobs
 
+  nightly:
```
coverbeck, line 91: The reason I created two workflows, instead of having just one controlled by the flag (my first attempt), is because if the `nightly` build runs while the `everything` build is running, the `everything` build will get cancelled.

.github/workflows/cron.yml
coverbeck, line 1: This is the GitHub actions file, removing it in favor of building on CircleCI.

.circleci/config.yml
```diff
@@ -30,60 +34,99 @@ common_filters: &common_filters
...
+              - build
+        - confidential-tool-tests:
+            <<: *common_filters
+            requires:
+              - build
+        - wes-toil-test:
+            <<: *common_filters
+            requires:
+              - build
+        - sonar-cloud:
```
coverbeck, line 71: I was torn on whether we should run SonarCloud on the nightly build against develop, but decided to leave it. Who knows, there might be some weird edge case where it reveals something.

denis-yuen:
> The trigger to fire this nightly is [here](https://app.circleci.com/settings/project/github/dockstore/dockstore-cli/triggers?return-to=https%3A%2F%2Fapp.circleci.com%2Fprojects%2Fproject-dashboard%2Fgithub%2Fdockstore%2F) -- will finalize it once this is merged.

neat

denis-yuen (approved):

.circleci/config.yml
```diff
@@ -30,60 +34,99 @@ common_filters: &common_filters
       ignore:
         - gh-pages
 
+common_jobs: &common_jobs
+    jobs:
+        - find-develop-snapshot
```
kathy-t, line 39: This is included in `common_jobs`, but inside `find-develop-snapshot`, it's only triggered when `condition: << pipeline.parameters.run_against_develop_core >>`. You could move this out of `common_jobs` and into `nightly` since the `everything` workflow doesn't use it. It would remove this "empty" job for `everything` https://app.circleci.com/pipelines/github/dockstore/dockstore-cli/1714/workflows/535c449e-f348-4ff2-b46a-6af475c0b7fa/jobs/8503

kathy-t (approved):

.circleci/config.yml
coverbeck: > This is included in `common_jobs`, but inside `find-develop-snapshot`, it's only triggered when `condition: << pipeline.parameters.run_against_develop_core >>`. You could move this out of `common_jobs` and into `nightly` since the `everything` workflow doesn't use it. It would remove this "empty" job for `everything` https://app.circleci.com/pipelines/github/dockstore/dockstore-cli/1714/workflows/535c449e-f348-4ff2-b46a-6af475c0b7fa/jobs/8503

I tried to address this, but you [can't share parts](https://stackoverflow.com/questions/9254178/is-there-yaml-syntax-for-sharing-part-of-a-list-or-map) of a YAML list via YAML anchors. In other words, I couldn't have the `find-develop-snapshot` job as part of the `nightly` build, then add the rest of the jobs via the YAML alias.

So there are two choices:

1. Leave it the way it is, with the empty job being "run" every time for regular builds, or
2. Get rid of `common_jobs` and mainly duplicate the list of jobs for both everything and nightly in the config.yml.

I don't really like either one of these approaches, but lean towards no duplication being the lesser evil.

@kathy-t, who raised the issue, and anybody else monitoring this, are you OK with that? I know you already approved it, but just wanted to make sure.

denis-yuen: > I don't really like either one of these approaches, but lean towards no duplication being the lesser evil. 

Agreed, looking superficially we get charged per minute and there doesn't seem to be a cost for container startup

kathy-t: I'm okay with keeping it the way you have in the PR. Could you add a comment about it? I feel like I would forget about why it's that way in the future and accidentally go down the route of trying to remove it :grin:

coverbeck:
Note: purposely merging this to develop instead of release branch, since it only affects the develop branch.

PR #269, opened 2023-12-06
finish deprecation of author in favour of authors array

denis-yuen:
**Description**
Supports https://github.com/dockstore/dockstore/pull/5706 
Adapts the CLI to accept the new authors list

**Review Instructions**
Build should pass mostly, bonus points for trying out the CLI list and update functionality in QA. 

**Issue**
https://github.com/dockstore/dockstore/issues/5467

**Security**


denis-yuen:
Supports https://github.com/dockstore/dockstore/pull/5706

pom.xml
```diff
@@ -38,7 +38,7 @@
 
         <github.url>scm:git:git@github.com:dockstore/dockstore-cli.git</github.url>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dockstore-core.version>1.15.0-alpha.11</dockstore-core.version>
```
denis-yuen, line 41: Will make a proper release after merge of main PR (or possibly during staging release tagging)

dockstore-client/generated/src/main/resources/pom.xml
```diff
@@ -30,7 +30,7 @@
     <dependency>
       <groupId>org.glassfish.jersey.inject</groupId>
       <artifactId>jersey-hk2</artifactId>
-      <version>3.0.10</version>
```
denis-yuen, line 33: Think the Dockstore core update pulled in some dependabot updates and whatnot

dockstore-client/src/main/java/io/dockstore/client/cli/nested/DepCommand.java
```diff
@@ -62,8 +62,10 @@ public static boolean handleDepCommand(String[] args) {
             ApiClient defaultApiClient;
             defaultApiClient = Configuration.getDefaultApiClient();
             MetadataApi metadataApi = new MetadataApi(defaultApiClient);
+            // todo: revert after real release
```
denis-yuen, line 65: Will need to revert this before merge, semver didn't like the tag name I picked

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -810,7 +811,7 @@ public void handleInfo(String entryPath) {
                     description = "";
                 }
 
-                String author = container.getAuthor();
+                String author = container.getAuthors().isEmpty() ? "" : container.getAuthors().get(0).getName();
```
coverbeck, line 814: If I'm reading this correctly, I didn't realize the CLI hadn't already been updated for the deprecation. Does this mean the 1.14 CLI will be broken against the 1.15 web service? If so, would it be better to not remove the deprecated fields from the web service until one more release?

coverbeck (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
denis-yuen: I think that yes, these two commands (updating and listing tools) will be broken. 
Technically, if we go one more release, I'd probably break apart the webservice PR into two parts 
1. deletion of the old authors fields in the models and DB can wait
2. for now, redirect the tests and queries

denis-yuen: thoughts @kathy-t ?

svonworl (approved):

kathy-t (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
denis-yuen: Had a chat, we decided to make the webservice backwards compatible for existing 1.14 CLI users but merge most of the ui and cli changes so we don't have this problem in the 1.17 release

denis-yuen:
Re-requesting due to https://github.com/dockstore/dockstore/pull/5706#issuecomment-1854677325 
i.e. this just relies on it being backwards compatible anyway

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -810,7 +811,7 @@ public void handleInfo(String entryPath) {
                     description = "";
                 }
 
-                String author = container.getAuthor();
```
coverbeck, line 813: Based on the discussion, weren't you going to change this and call `getAuthor()` after all?

kathy-t (approved):

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
denis-yuen: No, I thought the idea was that the 1.15 webservice would have both. 

We would make the 1.15 cli use 'authors' instead of author, but both the 1.14 and 1.15 cli would thus work with the 1.15 webservice. Then in 1.16 when we remove `author`, we're hoping that users would be using the 1.15 cli and thus not notice anything.

coverbeck (approved):

svonworl (approved):

PR #284, opened 2024-04-01
SEAB-6097: Nightly CLI build failure should notify in Slack

ll5zh:
**Description**
This PR restores the Slack notification for nightly CLI build failures, which was removed after moving the nightly cron to CircleCI (https://github.com/dockstore/dockstore-cli/pull/268). Notifications are set up appear in #dockstore-dev-alerts.

**Review Instructions**
In config.yml, verify that the Slack notification is configured as expected. See https://oicr.slack.com/archives/C0151UDNR9R/p1711479933594839 for a successful test run of the `slack/notify` setup (without the nightly condition).

Since the CLI nightly workflow consists of multiple jobs, the PR currently proposes separate Slack notifications for the build, unit tests, and integration tests. Are there any jobs that should be added/removed?

Note: I originally created a new Slack app called "Dockstore CLI Build", with its own CircleCI context called `oicr-slack-cli`, to do the notifying (test run: https://oicr.slack.com/archives/C0151UDNR9R/p1711465118013749). But it looks like the notification also works with our pre-existing "dockstore build notifications" app and `oicr-slack` context. I decided to stick with the old app for consistency, but do comment if you'd prefer otherwise.

**Issue**
SEAB-6097

**Security**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,8 @@ common_jobs: &common_jobs
             <<: *common_filters
             requires:
               - find-develop-snapshot
+            context:
+              - oicr-slack-cli
         - unit-tests:
             <<: *common_filters
             requires:
@@ -230,6 +232,13 @@ jobs:
           root: .
           paths:
             - .
+        # Remove nightly condition for now, testing app setup
+      - slack/notify:
+          event: fail
+          template: basic_fail_1
+      - slack/notify:
+          event: pass
+          template: basic_success_1
   sonar-cloud:
     docker: # run the steps with Docker
       - image: cimg/openjdk:<< pipeline.parameters.java-tag >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,9 @@ parameters:
   run_against_develop_core:
     type: boolean
     default: false
+  test_slack: # added temporarily for testing purposes
+    type: boolean
+    default: true
 
 orbs:
   build-tools: circleci/build-tools@2.7.0
@@ -35,6 +38,10 @@ common_filters: &common_filters
       ignore:
         - gh-pages
 
+slack_context: &slack_context
+  context:
+    - oicr-slack
+
 common_jobs: &common_jobs
     jobs:
         # find-develop-snapshot job is run for both "everything" and "nightly" builds even though it's only needed for nightly.
@@ -42,35 +49,35 @@ common_jobs: &common_jobs
         - find-develop-snapshot
         - build:
             <<: *common_filters
+            <<: *slack_context
             requires:
               - find-develop-snapshot
-            context:
-              - oicr-slack
         - unit-tests:
             <<: *common_filters
+            <<: *slack_context
             requires:
               - build
-            context:
-              - oicr-slack
         - integration-tests:
             matrix:
               parameters:
                 testing_profile: [ "singularity-tests", "bitbucket-tests" ]
             <<: *common_filters
+            <<: *slack_context
             requires:
               - build
-            context:
-              - oicr-slack
         - non-confidential-tests:
             <<: *common_filters
+            <<: *slack_context
             requires:
               - build
         - confidential-workflow-tests:
             <<: *common_filters
+            <<: *slack_context
             requires:
               - build
         - confidential-tool-tests:
             <<: *common_filters
+            <<: *slack_context
             requires:
               - build
         - wes-toil-test:
@@ -141,10 +148,7 @@ jobs:
       - save_test_results
       - send_coverage
       - persist_coverage
-      - when:
-          condition: << pipeline.parameters.run_against_develop_core >>
-          steps:
-            - notify-slack
+      - notify-slack
 
   non-confidential-tests:
     executor: machine_integration_test_exec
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -43,17 +44,23 @@ common_jobs: &common_jobs
             <<: *common_filters
             requires:
               - find-develop-snapshot
+            context:
```
kathy-t, line 47: > Since the CLI nightly workflow consists of multiple jobs, the PR currently proposes separate Slack notifications for the build, unit tests, and integration tests. Are there any jobs that should be added/removed?

GitHub won't let me comment on the lines I want, but is it possible to call `notify-slack` within `common_jobs`, after the `build` job? So the slack notification will alert us if the `nightly` workflow fails as a whole and we don't have to add the `notify-slack` command to multiple jobs

https://github.com/dockstore/dockstore-cli/blob/9cb046aa37932121d161d266cd66825f4e04eda1/.circleci/config.yml#L38-L43

coverbeck (commented):
Agreed with Kathy's feedback, nothing else to add. Please re-request review when you've responded to it, thanks.

.circleci/config.yml
ll5zh: From what I've seen, I don't think the `slack/notify` command can assess the cumulative status of multiple jobs in a workflow (see "Is it possible to receive only the final status of the Workflow?" in https://github.com/CircleCI-Public/slack-orb/wiki/FAQ).

Seeing if we can integrate Slack via webhook instead of the orb -- it looks like CircleCI has a `workflow-completed` event (https://circleci.com/docs/webhooks/#event-specifications) which could satisfy the suggested feedback.

kathy-t: Ah, I see, I wasn't sure if it was possible so you can ignore that feedback :slightly_smiling_face: 

Is it possible to add `notify-slack` to `setup_and_run_integration_tests`? It's the only command called by `non-confidential-tests`, `confidential-workflow-tests`, `confidential-tool-tests`, and `integration-tests`. That way you don't have to add it to each integration test job

https://github.com/dockstore/dockstore-cli/blob/9cb046aa37932121d161d266cd66825f4e04eda1/.circleci/config.yml#L550-L557

ll5zh: I think that would work, since `notify-slack` would be a step within a job 👍 It would prevent repetition in the config file, but I guess we'd still have separate notifications for each job (as opposed to one workflow notification).

Will give it a try first, so we can decide how we feel about the multiple notifications

denis-yuen: It does seem a bit excessive to report with separate Slack notifications for the build, unit tests, and integration tests. 
Could consider notifying after sonar cloud which should only pass if all the other builds pass.

ll5zh: A possible concern with only notifying in sonar cloud (if I'm understanding CircleCI correctly): if the build fails, the unit/integration tests (which require the build) won't run at all, and then sonar cloud (which requires those tests) won't run also? So we won't be notified at all.

To be fair, if we're only notifying on failures, the notifications shouldn't be a daily occurrence anyways

denis-yuen: Oh, I see. So you would be able to notify on success, but not on failure.  :thought_balloon:

.circleci/config.yml
```diff
@@ -536,6 +553,7 @@ commands:
       - save_test_results
       - send_coverage
       - persist_coverage
+      - notify-slack
```
ll5zh, line 556: For testing purposes, this is what the Slack notifications looked like when notified separately on each job's success: https://oicr.slack.com/archives/C0151UDNR9R/p1712088975956109

Things to consider:

- If we choose to notify Slack only on failed jobs, these messages probably won't occur too frequently; nightly runs will probably succeed most of the time
- I don't think it'd be possible to get failed notifications from the build, unit tests, and integration tests all at once. If the build fails/notifies, for example, the dependent jobs (i.e. all the tests) should be skipped and can't notify. Worst case scenario is if the build passes, and all the unit/integration tests fail?
- Also just noticed that some integration tests produce multiple notifications (e.g. 3 for confidential-tool/workflow-tests, 2 for non-confidential-tests); appears to be due to the parallelism

denis-yuen: > * If we choose to notify Slack only on failed jobs, these messages probably won't occur too frequently; nightly runs will probably succeed most of the time

I think notifying on failed jobs is better since those are the actionable ones. That said, we can run into the problem that it the notifications break ... 

Especially with the parallel notifications, 

> some integration tests produce multiple notifications 

I think there are too many notifications.


> Worst case scenario is if the build passes, and all the unit/integration tests fail?

If I understand correctly, we'd get messages for all the failures if we were notifying on failures only. 
I think that's ok. I think numerous failure messages in a worse case scenario is better than numerous success messages (which may lead us to ignore messages)

Open to ideas

coverbeck (approved):
> Oh, I see. So you would be able to notify on success, but not on failure. 💭
> 
> > To be fair, if we're only notifying on failures, the notifications shouldn't be a daily occurrence anyways
> 
> Ah good point, if this PR is notifying only on failure that would be a massive improvement in the first place.

Agreed it would be very nice if we only got one failure notification, but I don't see a way to do it.

kathy-t (approved):

denis-yuen (approved):

PR #293, opened 2024-08-13
Compatible with (but not mandatory) Java 21

denis-yuen:
**Description**
Make the CLI build compatible with Java 21.
Does not switch to only Java 21, yet

Some notes:

- A new version of Ubuntu so we get Java 21 with a new machine executor
- New Ubuntu means new Docker daemon version which finally terminates support for older Docker images https://docs.docker.com/engine/deprecated/#pushing-and-pulling-with-image-manifest-v2-schema-1
  - this means I needed to touch and update a bunch of old tutorial images
  - in some cases this means forking into `dockstore-testing`, in others just download and uploading with a  new daemon
  - this does mean that our tutorials may die over time with new Docker versions
- fixing failed tests is a pain after the death of System Stubs and its replacement with System Lambda i.e. we cannot catch System.exit as an exception anymore across the board, instead the whole VM exits with an unexpected exit with little output, interrupting test result upload (something linked to https://github.com/dockstore/dockstore/issues/5284 )
- some images forked and uploaded into `quay.io/dockstore-testing`, some workflows modified in testing repos such as https://github.com/dockstore-testing/dockstore-tool-md5sum https://github.com/dockstore-testing/mutect-tool https://github.com/dockstore-testing/dockstore-tool-linux-sort but in the end, I mostly downloaded and uploaded Docker images using a compatible Docker daemon rather than fixing old Dockerfiles
- there's a complex set of four mocked tests using powermock (which is dead) and  old versions of junit and mockito. While useful, on balance they probably are not critical or worth blocking testing our compatibility with Java 21, so splitting up a follow-up ticket https://ucsc-cgl.atlassian.net/browse/SEAB-6851

**Review Instructions**
Could do some spot checking, listing and running workflows as in our getting started tutorials

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6611
Follow-up with https://ucsc-cgl.atlassian.net/browse/SEAB-6851

**Security**


denis-yuen:
To write up my findings before the weekend

1. New Ubuntu to get a new Java version
2. New Ubuntu means new Docker (which means many of our oldest Docker images are no longer compatible due to https://github.com/dockstore/dockstore/issues/5878 )
3. New pipx install method seemed to mean that I needed to install schema salad, cwlref-runner and their paths separately

scripts/install-tests.sh
```diff
@@ -14,7 +14,15 @@ fi
 if [ "${TESTING_PROFILE}" = "toil-integration-tests" ]; then
     pip3 install --user toil[cwl]==7.0.0
 else
-    pip3 install --user -r https://raw.githubusercontent.com/dockstore/dockstore/develop/dockstore-webservice/src/main/resources/requirements/1.15.0/requirements3.txt
+    sudo apt-get update
+    # https://stackoverflow.com/questions/44331836/apt-get-install-tzdata-noninteractive needed by cwltool
```
denis-yuen, line 18: This matches 1.16 quickstart with pipx, but had to do a bit of extra work to get schema_salad and cwlref-runner (may be using pipx runpip wrong)

dockstore-cli-integration-testing/pom.xml
```diff
@@ -268,6 +268,8 @@
                         --add-opens java.base/java.base=ALL-UNNAMED
                         --add-opens java.base/java.lang.reflect=ALL-UNNAMED
                         --add-opens java.base/sun.nio.fs=ALL-UNNAMED
+                        <!-- for  Java 21, see   -->
```
coverbeck, line 271: Is this comment incomplete? See what?

dockstore-client/pom.xml
```diff
@@ -529,6 +529,8 @@
                         --add-opens java.base/java.io=ALL-UNNAMED
                         --add-opens java.base/java.util=ALL-UNNAMED
                         --add-opens java.base/java.base=ALL-UNNAMED
+                        <!-- for  Java 21, see   -->
```
coverbeck, line 532: Incomplete comment?

coverbeck (approved):

dockstore-cli-integration-testing/pom.xml
denis-yuen: extended comment, needed for system stubs in Java 21

dockstore-client/pom.xml
denis-yuen: extended comment, needed for system stubs in Java 21

.circleci/config.yml
```diff
@@ -17,14 +17,14 @@ executors:
 
   machine_integration_test_exec:
     machine: # run the steps with Ubuntu VM
-      image: ubuntu-2204:2023.10.1
+      image: ubuntu-2204:2024.04.4
```
kathy-t, line 20: Why is the tag different for this and the `toil_wes_test_executor` below?

kathy-t (approved):

dockstore-client/src/test/resources/testDirectory3/mutect.cwl
```diff
@@ -10,7 +10,7 @@ doc: "Mutect 1.1.5"
 
 hints:
   - class: DockerRequirement
-    dockerPull: quay.io/jeltje/mutect
```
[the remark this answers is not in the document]
svonworl: Don't understand why we deleted the `dockerPull`, can you please elaborate?

.circleci/config.yml
denis-yuen: A sad reason
While experimenting, I bumped to a newer version to try to get a newer Java version on the machine executor. 
It turned out to be unnecessary, but since we ran the tests anyway I kept the newer versions that worked. 

Something about `ubuntu-2204:2024.11.1` breaks the toil install silently.

svonworl (approved):

PR #304, opened 2025-01-29
test out update

denis-yuen:
**Description**
A description of the PR, should include a decent explanation as to why this change was needed and a decent explanation as to what this change does

**Review Instructions**
Describe if this ticket needs review and if so, how one may go about it in qa and/or staging environments.
For example, a ticket based on Security Hub, Snyk, or Dependabot may not need review since those services
will generate new warnings if the issue has not been resolved properly. On the other hand, an infrastructure
ticket that results in visible changes to the end-user will definitely require review.
Many tickets will likely be between these two extremes, so some judgement may be required.

**Issue**
A link to a github issue or SEAB- ticket (using that as a prefix)

**Security**

- [ ] Check that you pass the basic style checks and unit tests by running `./mvnw clean install` 
- [ ] If this PR is for a user-facing feature, create and link a documentation ticket for this feature (usually in the same milestone as the linked issue). Style points if you create a documentation PR directly and link that instead.


denis-yuen:
Seems unrelated, the PGP server for `com.auth0:java-jwt:jar:3.8.1` may have gone under

dockstore-cli-integration-testing/generated/src/main/resources/pom.xml
```diff
@@ -49,13 +49,19 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>dockstore-common</artifactId>
-      <version>1.17.0-alpha.0</version>
+      <version>1.17.0-alpha.2</version>
       <scope>compile</scope>
+      <exclusions>
```
denis-yuen, line 54: Note: this library is used by the webservice when it uses a library (org.kohsuke github-api) It can safely be excluded by the CLI (confirmed by these tests)
I do not know why the webservice build has not died for the same reason but it it does, you can exclude the key in the keylist file as we do with the jackson libraries. 

We could also try upgrading as noted by 
https://github.com/auth0/java-jwt

svonworl (approved):

kathy-t (approved):

PR #310, opened 2025-07-11
Fix nightly CLI build

kathy-t:
**Description**
This PR fixes the nightly CLI build that has been failing (can see the failed message in dockstore-dev-alerts). Fixed it by creating a new 1.18.0-alpha.1 webservice tag and removing specific versions of some dependencies that are already specified in the bom-internal. See passing nightly build here https://app.circleci.com/pipelines/github/dockstore/dockstore-cli/2324/workflows/28b21063-0847-4773-a3b4-907b02939f9c

**Review Instructions**
Nightly CLI build should pass

**Issue**
n/a

**Security**


dockstore-client/pom.xml
```diff
@@ -276,18 +274,6 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-vfs2</artifactId>
-            <!-- remove after bom is updated with Dockstore or above 1.14.0 -->
-            <version>2.10.0</version>
-            <exclusions>
```
kathy-t, line 281: It's unclear to me if these exclusions should be kept

kathy-t (commented):
Hmm guess I can't submit my own review comment without having something here

dockstore-client/pom.xml
denis-yuen: I think `commons-httpclient`, you can try building without it (I think this was a crude way of managing which versions we got). 

I think `hadoop-hdfs-client` can stay excluded (I'm pretty sure we'll never use it, it's integration with the hadoop file system)

denis-yuen (approved):

svonworl (approved):

